PR #2, opened 2019-03-12
rename to make room for other roles

fromanirh:
Up until now the SSP operator had just one task:
install the common templates.
That lead to quite some simplifications in the project layout.

We want the operator to deploy other SSP components, so
we make room for other roles.


fromanirh:
@MarSik this is an (educated) search/replace, not sure I made everything properly. Please check

deploy/crds/kubevirt_v1_commontemplatesbundle_cr.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kubevirt.io/v1
+kind: KubevirtCommonTemplateBundle
```
MarSik: You use KubevirtCommonTermplate_s_Bundle in the CRD.

fromanirh:
> It looks good, the OLM files can be re-generated using one of the operator-sdk commands.
Do you remember which command? otherwise I'll look it up in the docs.

Doing some manual testing now.

ksimon1:
lgtm, I tested it with my PR https://github.com/MarSik/kubevirt-ssp-operator/pull/3 and it works as expected

PR #4, opened 2019-03-15
validator: add role and helpers to install

fromanirh:
Add support to deploy the template-validator

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #6, opened 2019-03-22
validator: add role to install

fromanirh:
Add support to deploy the template validator, using
the https://github.com/openshift/service-serving-cert-signer
facility provided by OKD/OCP

Replaces https://github.com/MarSik/kubevirt-ssp-operator/pull/4


fromanirh:
Needs to be tested. On it.

build/Dockerfile
```diff
@@ -1,5 +1,11 @@
 FROM quay.io/water-hole/ansible-operator
 
+USER 0
+COPY repos/RPM-GPG-KEY-CentOS-SIG-PaaS /etc/pki/rpm-gpg
+COPY repos/centos-openshift-origin.repo /etc/yum.repos.d
+RUN yum install -y openssl origin-clients
+USER 1001
```
fromanirh: Uhm, we don't actually need this (and related repo files), cleaning up

fromanirh:
seems to work. I want to do another verification e2e with updated common-templates, but we can merge this PR

roles/KubevirtTemplateValidator/files/service.yaml
```diff
@@ -0,0 +1,89 @@
...
+  template:
+    metadata:
+      name: virt-template-validator
+      labels:
+        kubevirt.io: virt-template-validator
+    spec:
+      serviceAccountName: template-validator
+      containers:
+        - name: webhook
+          image: quay.io/fromani/kubevirt-template-validator:v0.3.0
```
MarSik: Can we extract this to a variable (role default)? Just so we can make downstream builds easier.

PR #9, opened 2019-03-29
add preprocessing script

ksimon1:
This PR adds preprocessing script which updates common templates with another annotation.
Script has 2 parameters: patchFilePath, commonTemplatesFolderPath, 
```e.g. python ${HOME}/preprocessing.py ${HOME}/patch.yaml ${HOME}/roles/KubevirtCommonTemplatesBundle/files/ ```
it iterates over each file in commonTemplatesFolderPath and updates templates.

Patch file should have format as mentioned in patch.yaml file
When it matches label, it will add annotation.

### CODE REVIEW

File: build/Dockerfile [hunk 1]
```diff
@@ -3,3 +3,6 @@ FROM quay.io/water-hole/ansible-operator
 COPY roles/ ${HOME}/roles/
 COPY playbooks/ ${HOME}/
 COPY watches.yaml ${HOME}/watches.yaml
+COPY preprocessing.py ${HOME}/
+COPY patch.yaml ${HOME}/
+RUN python ${HOME}/preprocessing.py ${HOME}/patch.yaml ${HOME}/roles/KubevirtCommonTemplatesBundle/files/common-templates-v0.4.1.yaml ${HOME}/roles/KubevirtCommonTemplatesBundle/files/common-templates-v0.4.1.yaml
```
MarSik: Actually, just iterate over all common templates versions you will find.

build/preprocessing.py
```diff
@@ -0,0 +1,55 @@
...
+
+  for template in commonTemplates:
+    metadata = template.get("metadata", {})
+    annotations = metadata.get("annotations",{})
+    labels = metadata.get("labels", {})
+    for rule in rules:
+      key = rule.get("matchLabel")[0]
+      value = rule.get("matchLabel")[1]
+      if labels.get(key) == value:
+        annotationKey = rule.get("addAnnotation")[0]
```
MarSik: Please add a debug print here that prints the template name and the added annotation

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
+#! python
```
fromanirh: I'd rename like "preprocess_template.py" or "add_labels" or something more expressive.

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+
+import yaml
+import sys
+from os import listdir
+from os.path import isfile, join
+
+if __name__ == "__main__":
```
fromanirh: better pack all the application code in a function, and call it from here.

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+      
+      rules = patch.get("rules")
+      for rule in rules:
+        rule["matchLabel"] = rule.get("matchLabel").split(": ")
+        rule["addAnnotation"] = rule.get("addAnnotation").split(": ")
+    except yaml.YAMLError as exc:
+      print(exc)
+      sys.exit(1)
+
+  commonTemplatesPath = sys.argv[2] if len(sys.argv) > 2 else "/opt/ansible/roles/KubevirtCommonTemplatesBundle/files"
```
fromanirh: please use constants (well, actually module-level globals) to make evident the defaults we use.

fromanirh: Also, do we want to use something like the argparse module?

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
+#! python
+
+import yaml
+import sys
+from os import listdir
+from os.path import isfile, join
```
fromanirh: injecting content from modules in the global namespace is bad practice in general.
In this case the script is short enough to be somehow OK.

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+      labels = metadata.get("labels", {})
+      annotationsAdded = []
+      for rule in rules:
+        key = rule.get("matchLabel")[0]
+        value = rule.get("matchLabel")[1]
+        if labels.get(key) == value:
+          annotationKey = rule.get("addAnnotation")[0]
+          annotationValue = rule.get("addAnnotation")[1]
+          annotations[annotationKey] = annotationValue
+          annotationsAdded.append(annotationKey + ": " +  annotationValue)
```
fromanirh: we need some tests for this code. It can be done in a followup patch if needed.

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+          print("Nothing in common templates file")
+          continue
+
+      except yaml.YAMLError as exc:
+        print(exc)
+        continue
+
+    print("------------------------------------------------------")
+    print("Running script for file: " + fileName)
+    print("------------------------------------------------------")
```
fromanirh: the processing steps are well defined enough to fit each in a separate function. This would be a nice improvement, but the biggest gain is to move all the application code in a function, to make it easily testable.

build/preprocessing.py
```diff
@@ -0,0 +1,72 @@
...
+        if labels.get(key) == value:
+          annotationKey = rule.get("addAnnotation")[0]
+          annotationValue = rule.get("addAnnotation")[1]
+          annotations[annotationKey] = annotationValue
+          annotationsAdded.append(annotationKey + ": " +  annotationValue)
+
+      if len(annotationsAdded) > 0:
+        print("Updating "+ metadata.get("name")+ " template")
+        for a in annotationsAdded:
+          print("adding " + a)
```
fromanirh: the logging module is better than plain print()s, but this improvement could be done later.

build/preprocessing.py
MarSik: I do not expect any other arguments so it might be a bit of an overkill.

build/preprocessing.py
MarSik: Logging is cool, but since the script is supposed to be used from docker build it is a bit of an overkill as well. But Karel already did it...

ksimon1: i can change it back, it is easy change.

fromanirh: @ksimon1 don't bother: let's focus on add basic testing to make sure the script works: troubleshooting issues during docker builds is 10x (at least) more painful

MarSik: @ksimon1 Yep, no need to rewrite it back, logger is better and we need to focus on other stuff before fixing nice to haves.

build/Dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM quay.io/water-hole/ansible-operator
 
+COPY preprocess_template.py ${HOME}/
+COPY test_preprocess_template.py ${HOME}/
+RUN python ${HOME}/test_preprocess_template.py
```
fromanirh: Uhm, why? We want to have tests and run them ourselves when developing, not in the image I believe.

fromanirh:
I think we should not run the tests inside the docker image - just manually when developing. Besides that looks OK. How did you tested this PR?

build/Dockerfile
ksimon1: If it is only for developing purpose, then i will delete it

ksimon1:
@fromanirh 
1) Update patch.yaml in deploy folder with e.g.:
```
rules:
  - matchLabel: "os.template.cnv.io/win2k12r2: true"
    addAnnotation: "myAwesomeAnnotation: true"
```
2) build image
It should log e.g. something like:
```
INFO:root:Updating win2k12r2-generic-medium template
INFO:root:adding myAwesomeAnnotation: true
INFO:root:Updating win2k12r2-generic-large template
INFO:root:adding myAwesomeAnnotation: true
```
3) deploy ssp-operator into cluster
4) wait few second until templates are deployed
5) oc get templates win2k12r2-generic-large -o yaml | grep myAwesomeAnnotation

fromanirh:
LGTM. Do we want to move patch.yaml in a configmap?

PR #13, opened 2019-04-09
namespace: make it configurable

fromanirh:
address possible namespace variable clash, make the namespace configurable using a variable


fromanirh:
@MarSik tested commit https://github.com/MarSik/kubevirt-ssp-operator/commit/81b5010f134e85c151d206bc673c06b7ebcebaeb - the operator deploys the validator fine. Didn't test changing the default namespace, though.
I guess it's OK, Installing the validator in a ns != kubevirt was never really tested.

PR #15, opened 2019-04-10
add patchField feature into preprocessing script

ksimon1:
This PR adds possibility to patch update fields in common templates.
patch.yaml file was updated with examples how to use this new feature.
@fromanirh, @MarSik  please review

### DIFF
--- a/build/patch.yaml
+++ b/build/patch.yaml
@@ -1,3 +1,22 @@
 #rules:
-#  - matchLabel: "os.template.cnv.io/win2k12r2: true"
-#    addAnnotation: "myAwesomeAnnotation: true"
+#  addAnnotation:
+#    - matchLabel: "os.template.cnv.io/centos7.0: true"
+#      addAnnotation: "myAwesomeAnnotation: true"
+#
+#  patchField rules replace value in specField attribute
+#  it can replace e.g. bool value, string, int, structures, ...
+#  patchField: 
+#    - matchLabel: "os.template.cnv.io/centos7.0: true" 
+#      specField: "objects.0.spec.template.spec.domain.cpu"
+#      value: 
+#        sockets: 1
+#        threads: 1
+#        cpu: 2
+#
+#    - matchLabel: "os.template.cnv.io/centos7.0: true" 
+#      specField: "objects.0.spec.template.spec.domain.cpu.sockets"
+#      value: 3
+#
+#    - matchLabel: "os.template.cnv.io/centos7.0: true" 
+#      specField: "objects.0.spec.running"
+#      value: true
--- a/build/preprocess_template.py
+++ b/build/preprocess_template.py
@@ -55,7 +55,7 @@ def process_rules(commonTemplates, rules):
             annotationKey = rule.get("addAnnotation")[0]
             annotationValue = rule.get("addAnnotation")[1]
             annotations[annotationKey] = annotationValue
-            annotationsAdded.append(annotationKey + ": " +  annotationValue)
+            annotationsAdded.append("adding " + annotationKey + ": " +  annotationValue)
 
       fieldsUpdated = []
       patchFieldRules = rules.get("patchField")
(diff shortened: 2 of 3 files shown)


build/preprocess_template.py
```diff
@@ -34,25 +40,53 @@ def load_common_templates(path):
...
+            for index, path in enumerate(specFieldPath):
+              if (index+1) == len(specFieldPath):
+                lastPathPiece = path
+                break
+              if type(obj) == list:
+                obj = obj[int(path)]
+              else:
+                obj = obj.setdefault(path, {})
+            obj[lastPathPiece] = rule.get("value")
+            fieldsUpdated.append("field " + ".".join(specFieldPath) + " updated with value: " + str(rule.get("value")))
```
fromanirh: we need to spend some time in refactoring once we reach a stable state

build/preprocess_template.py
```diff
@@ -34,25 +40,53 @@ def load_common_templates(path):
...
+      addAnnotationRules = rules.get("addAnnotation")
+      if addAnnotationRules != None:
+        for rule in addAnnotationRules:
+          key = rule.get("matchLabel")[0]
+          value = rule.get("matchLabel")[1]
+          if labels.get(key) == value:
+            annotationKey = rule.get("addAnnotation")[0]
+            annotationValue = rule.get("addAnnotation")[1]
+            annotations[annotationKey] = annotationValue
+            annotationsAdded.append(annotationKey + ": " +  annotationValue)
```
fromanirh: do we use "annotationsAdded" just for logging? If so:

  annotationsAdded.append("adding " + annotationKey + ": " + annotationValue)

build/preprocess_template.py
```diff
@@ -34,25 +40,53 @@ def load_common_templates(path):
...
+      if len(annotationsAdded) > 0 or len(fieldsUpdated) > 0:
         logging.info("Updating "+ metadata.get("name")+ " template")
-        for a in annotationsAdded:
-          logging.info("adding " + a)
+        if len(annotationsAdded) > 0:
+          for a in annotationsAdded:
+            logging.info("adding " + a)
+        if len(fieldsUpdated) > 0:
+          for p in fieldsUpdated:
+            logging.info(p)
```
fromanirh: let's simplify:

for msg in annotationsAdded + fieldsUpdated:
   logging.info(msg)

build/preprocess_template.py
```diff
@@ -70,7 +104,11 @@ def process_common_templates(commonTemplatesPath, rules):
     logging.info("Running script for file: " + fileName)
     logging.info("------------------------------------------------------")
 
-    updatedCommonTemplates = process_annotations(commonTemplates, rules)
+    updatedCommonTemplates = None
+    try:
+      updatedCommonTemplates = process_rules(commonTemplates, rules)
+    except Exception as e:
+      raise(e)
```
fromanirh: You use this pattern already, but re-raising the exception we catch without any action in between is just useless.
We can fix this later however.

build/patch.yaml
```diff
@@ -1,3 +1,22 @@
...
-#  - matchLabel: "os.template.cnv.io/win2k12r2: true"
-#    addAnnotation: "myAwesomeAnnotation: true"
+#  addAnnotation:
+#    - matchLabel: "os.template.cnv.io/centos7.0: true"
+#      addAnnotation: "myAwesomeAnnotation: true"
+#
+#  patchField rules replace value in specField attribute
+#  it can replace e.g. bool value, string, int, structures, ...
+#  patchField: 
+#    - matchLabel: "os.template.cnv.io/centos7.0: true"
```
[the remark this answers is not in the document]
MarSik: @fromanirh we can only use what is available in the base image. That limits us to pure Python + ansible dependencies. So yes, this is intentional.

build/preprocess_template.py
```diff
@@ -8,15 +8,21 @@
...
-        rule["addAnnotation"] = rule.get("addAnnotation").split(": ")
+      if rules.get("addAnnotation") != None:
+        for rule in rules.get("addAnnotation"):
+          rule["matchLabel"] = rule.get("matchLabel").split(": ")
+          rule["addAnnotation"] = rule.get("addAnnotation").split(": ")
+
+      if rules.get("patchField") != None:
+        for rule in rules.get("patchField"):
+          rule["matchLabel"] = rule.get("matchLabel").split(": ")
+          rule["specField"] = rule.get("specField").split(".")
```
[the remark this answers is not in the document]
MarSik: This is fine goal, but we needed something fast and it is supposed to be used just during operator build time. We are not writing a generic tool.

build/preprocess_template.py
MarSik: I agree.

build/preprocess_template.py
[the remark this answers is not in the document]
fromanirh: I guess that's good enough

build/preprocess_template.py
ksimon1: i will update it

PR #16, opened 2019-04-10
Make it easy to customize image names and other defaults

MarSik:
This change provides a common pattern for downstream customization. Just override the _defaults.yml file and set new values for the various role defaults.


MarSik:
@fromanirh @yanirq @ksimon1 Please check this out. We need a simple way to customize downstream packaging and this pattern might work.

playbooks/_defaults.yml
```diff
@@ -0,0 +1,4 @@
+# This file can be used to override default
+# role values. This is useful for "downstream"
+# packaging and opinionated releases.
+---
```
fromanirh: is not clear to me why we need both build/_defaults.yml and playbooks/_defaults.yml.

build/Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM quay.io/water-hole/ansible-operator
 COPY roles/ ${HOME}/roles/
 COPY playbooks/ ${HOME}/
 COPY watches.yaml ${HOME}/watches.yaml
+COPY _defaults.yml ${HOME}/_defaults.yml
```
fromanirh: don't we need to use build/_defaults.yml here?

playbooks/_defaults.yml
MarSik: playbooks/_defaults.yml is empty by default and is there just to make sure the vars_files will not fail. The build one is the one supposed to replace the empty playbooks/ one with actual content.

build/Dockerfile
MarSik: You might be right. I considered the build directory to be the docker context for builds, but I am not exactly sure how the operator-sdk build works. I will test that tomorrow.

yanirq: I believe that everything invoked from the playbook should exist in the container. 
Maybe the build folder is not the best place for it logically (by operator-sdk semantics)

playbooks/_defaults.yml
yanirq: Can you add at least as a comment what could be the variables with examples maybe ?

build/Dockerfile
MarSik: Sure, that is what the COPY does. Maybe I should just move the defaults to the root directory?

MarSik: But we have the patch.yml and preprocessing script in build/ as well. And those are used for the same purpose (downstream customization)

yanirq: As long as they are copied i dont see a real problem

PR #18, opened 2019-04-11
validator: use version from CR

fromanirh:
previously, the field was just unused


fromanirh:
/assign @MarSik 
/cc @ksimon1 
Do we want to do something similar for the other components? it seems the common-templates can use a change like this

fromanirh:
tested on my cluster editing the CR to deploy validator v0.4.0 (v0.4.0 is still under testing) - works as expected

PR #19, opened 2019-04-12
validator: infer image repo from operator image

fromanirh:

### DIFF
--- a/playbooks/commontemplatesbundle.yaml
+++ b/playbooks/commontemplatesbundle.yaml
@@ -1,9 +1,5 @@
 - hosts: localhost
   gather_facts: no
-  tasks:
-  - name: Gather informations about the repository
-    import_role:
-      name: "KubevirtRepoInfo"
-  - name: Deploy the Common Templates
-    import_role:
-      name: "KubevirtCommonTemplatesBundle"
+  roles:
+  - KubevirtRepoInfo
+  - KubevirtCommonTemplatesBundle
--- a/playbooks/commontemplatesbundle.yaml
+++ b/playbooks/commontemplatesbundle.yaml
@@ -1,5 +1,9 @@
 - hosts: localhost
   gather_facts: no
   tasks:
-  - import_role:
+  - name: Gather informations about the repository
+    import_role:
+      name: "KubevirtRepoInfo"
+  - name: Deploy the Common Templates
+    import_role:
       name: "KubevirtCommonTemplatesBundle"
--- a/playbooks/commontemplatesbundle.yaml
+++ b/playbooks/commontemplatesbundle.yaml
@@ -1,5 +1,8 @@
 - hosts: localhost
   gather_facts: no
   roles:
-  - KubevirtRepoInfo
+# we want to prepare everything to pull containers from the same repo
+# as the operator, but we are not there just yet, so we disable the role
+# and we rely on backward compatible defaults.
+#  - KubevirtRepoInfo
   - KubevirtCommonTemplatesBundle
--- a/roles/KubevirtNodeLabeller/defaults/main.yml
+++ b/roles/KubevirtNodeLabeller/defaults/main.yml
@@ -2,8 +2,12 @@
 # defaults file for KubevirtNodeLabeller
 wait: true
 
-kubevirt_node_labeller: "quay.io/ksimon/kubevirt-node-labeller:v0.0.5"
-kvm_info_nfd_plugin: "quay.io/fromani/kvm-info-nfd-plugin:v0.4.0"
-kubevirt_cpu_nfd_plugin: "quay.io/ksimon/kubevirt-cpu-nfd-plugin:v0.0.4"
-libvirt_image: "kubevirt/virt-launcher:v0.16.0"
+kubevirt_node_labeller_image: "kubevirt-node-labeller"
+kubevirt_node_labeller_tag: "v0.0.5"
+kvm_info_nfd_plugin_image: "kvm-info-nfd-plugin"
+kvm_info_nfd_plugin_tag: "v0.4.0"
+kubevirt_cpu_nfd_plugin_image: "kubevirt-cpu-nfd-plugin"
+kubevirt_cpu_nfd_plugin_tag: "v0.0.4"
+libvirt_image: "virt-launcher"
+libvirt_tag: "v0.16.0"
 
--- a/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds-v0.0.5.yaml
+++ b/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds-v0.0.5.yaml
@@ -18,12 +18,12 @@ spec:
       serviceAccount: kubevirt-node-labeller
       containers:
       - name: kubevirt-node-labeller-sleeper
-        image: {{ kubevirt_node_labeller }}
+        image: "{{ repo_info['registry'] + '/' + kubevirt_node_labeller_image + ':' + kubevirt_node_labeller_tag }}"
         command: ["sleep"]
         args: ["infinity"]
       initContainers:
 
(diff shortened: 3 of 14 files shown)


roles/KubevirtTemplateValidator/tasks/main.yml
```diff
@@ -1,5 +1,18 @@
 ---
 # tasks file for KubevirtTemplateValidator
+- name: Extract the operator POD info
```
MarSik: Could you please extract this to a "playbook" and set_fact the operator_registry from it. That way all the other playbooks can use include_tasks to get this functionality and then easily use just the variable.

fromanirh: no problem, I was just unsure about the right way to share code. Let's try.

MarSik: The other option (maybe even better) might be to make this into a role.

roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -2,8 +2,12 @@
...
-kubevirt_cpu_nfd_plugin: "quay.io/ksimon/kubevirt-cpu-nfd-plugin:v0.0.4"
-libvirt_image: "kubevirt/virt-launcher:v0.16.0"
+kubevirt_node_labeller_image: "kubevirt-node-labeller"
+kubevirt_node_labeller_tag: "v0.0.5"
+kvm_info_nfd_plugin_image: "kvm-info-nfd-plugin"
+kvm_info_nfd_plugin_tag: "v0.4.0"
+kubevirt_cpu_nfd_plugin_image: "kubevirt-cpu-nfd-plugin"
+kubevirt_cpu_nfd_plugin_tag: "v0.0.4"
+libvirt_image: "virt-launcher"
+libvirt_tag: "v0.16.0"
```
fromanirh: please note this change breaks the Labeller completely because it is using images from 3 sources:
- quay.io/ksimon
- quay.io/fromani
- kubevirt

Also, this assumes the container name is stable.

playbooks/commontemplatesbundle.yaml
```diff
@@ -1,5 +1,9 @@
 - hosts: localhost
   gather_facts: no
   tasks:
-  - import_role:
+  - name: Gather informations about the repository
```
MarSik: I think you can also use just:
roles:
- KubevirtRepoInfo
- KubevirtCommonTemplatesBundle

See https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-duplication-and-execution

roles/KubevirtNodeLabeller/defaults/main.yml
MarSik: Right, I think we can assume the name is stable. Also we could use defaults to only use the detected name if the role was executed. And that we could do just for downstream.

playbooks/commontemplatesbundle.yaml
fromanirh: Thanks, will fix.

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds-v0.0.5.yaml
```diff
@@ -18,12 +18,12 @@ spec:
       serviceAccount: kubevirt-node-labeller
       containers:
       - name: kubevirt-node-labeller-sleeper
-        image: {{ kubevirt_node_labeller }}
+        image: "{{ repo_info['registry'] + '/' + kubevirt_node_labeller_image + ':' + kubevirt_node_labeller_tag }}"
```
MarSik: Please create an ssp_registry variable intead of using the ["registry"] references. Then you could do for example `{{ ssp_registry |default("kubevirt") }}/{{ image }}:{{ tag }}` and kubevirt will be used for upstream where the role was disabled (for example)

roles/KubevirtRepoInfo/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+    kind: Pod
+    label_selectors:
+      - name = kubevirt-ssp-operator
+  register: operator_pod
+- name: Extract the POD registry
+  set_fact:
+    operator_image_name: "{{ operator_pod['spec']['containers'][0]['image'] }}"
+- name: parse image info
+  set_fact:
+    repo_info: "{{ dict(['registry', 'image'] | zip(operator_image_name.rsplit('/', 1))) }}"
```
MarSik: As I said above, please create a fact that only contains the registry to make it easy to use.

roles/KubevirtNodeLabeller/defaults/main.yml
ksimon1: after labeller, cpu plugin, kvm info will be transfered under kubevirt organisation, it will be only one source(i am not sure about virt-launcher).

fromanirh: @ksimon1 sure, once that is done it will be right (even virt-launcher). But atm is broken :\

roles/KubevirtTemplateValidator/tasks/main.yml
fromanirh: went for making this into a role

roles/KubevirtRepoInfo/tasks/main.yml
fromanirh: Done

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds-v0.0.5.yaml
fromanirh: Good idea, done - also set backward-compatible defaults for all images

roles/KubevirtRepoInfo/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: extract the repo into a variable
```
MarSik: I think you can set multiple facts in one task.

playbooks/commontemplatesbundle.yaml
```diff
@@ -1,5 +1,5 @@
 - hosts: localhost
   gather_facts: no
-  tasks:
-  - import_role:
-      name: "KubevirtCommonTemplatesBundle"
+  roles:
+  - KubevirtRepoInfo
```
MarSik: The upstream version should document this, but we must keep the role disabled to make sure the defaults for NFD containers are used

fromanirh: OK. What'ts the best way to disable the role?

MarSik: Hmm I would say comment it out with a proper explanation :)

fromanirh: Ack

roles/KubevirtRepoInfo/tasks/main.yml
fromanirh: Done, need to test this

fromanirh:
@MarSik turns out I can't really consolidate the set_facts, otherwise I get "undefined variable" errors. I think is not that big of a deal.
Tested the code, works, but please check the caveat noted in https://github.com/MarSik/kubevirt-ssp-operator/pull/19/commits/27615f11fb044e40d1ecbba82f001059b7d765aa

roles/KubevirtRepoInfo/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
+  author: your name
```
MarSik: We should probably fix or remove the meta for this role.

fromanirh: Also for the other roles - they are all meant to be used internally, and not published. Will fix in a new PR.

PR #21, opened 2019-04-15
Go api

fromanirh:
add GO packages to integrate with HCO.
go structs defiition belongs there, putting them in other projects which need them (e.g. HCO) or in different repos will make harder to keep them in sync.

In the long run, we should actually (re)generate the CR(D?)s from the GO code, in order to have just one source of truth.
Doing the reverse (generate go code from YAML defs) is possibile but AFAIK harder because the lack of tooling.


fromanirh:
We can merge this. It doesn't affect existing functionality, it is self-contained and testing without this patch is much, much harder.

fromanirh:
obsoleted by https://github.com/MarSik/kubevirt-ssp-operator/pull/23

PR #23, opened 2019-04-16
Go api - try #2

fromanirh:
add GO packages to integrate with HCO.
go structs definition belongs there, putting them in other projects which need them (e.g. HCO) or in different repos will make harder to keep them in sync.

In the long run, we should actually (re)generate the CR(D?)s from the GO code, in order to have just one source of truth.
Doing the reverse (generate go code from YAML defs) is possibile but AFAIK harder because the lack of tooling.

Worth mentioning that another option is enhance the HCO to read YAML CR specs from files, but this requires longer discussion, which has not begun yet, and probably synchronization with other teams.


MarSik:
Yeah, I would very much prefer if HCO was not forcing other components to use specific style and read just yamls. But we need something right now so there is no avoiding this atm. Fortunately this does not interfere with the rest of the files.

PR #25, opened 2019-04-16
Expose more variables to _defaults.yml

MarSik:
We need to expose the defaults even when the value is then provided by the CR. Both common templates and the validator are used like that.


MarSik:
@fromanirh Please check this out, I think I need it for downstream

PR #27, opened 2019-04-17
rbac: manifests: prune roles

fromanirh:
per https://kubernetes.io/docs/reference/access-authn-authz/rbac/:

"""
In the RBAC API, a role contains rules that represent a set of permissions.
Permissions are purely additive (there are no “deny” rules).
A role can be defined within a namespace with a Role, or cluster-wide with
a ClusterRole.
[...]
A ClusterRole can be used to grant the same permissions as a Role,
but because they are cluster-scoped, they can also be used to grant
access to:

- cluster-scoped resources (like nodes)
- non-resource endpoints (like “/healthz”)
- namespaced resources (like pods) across all namespaces (needed
  to run kubectl get pods --all-namespaces, for example)
"""

So we actually need only ClusterRoles, and their binding.


MarSik:
Lgtm, but please make sure it works. This was the trickiest part of the whole operator stuff.

fromanirh:
> Lgtm, but please make sure it works. This was the trickiest part of the whole operator stuff.

Sure, will test and report. Please not merge until then.

fromanirh:
the template-validator, which was the component pulling these deps, deployed fine with this patch applied. However, I found https://github.com/operator-framework/operator-sdk/issues/1380 - which I think is unrelated. I think we can merge this patch after rebase

fromanirh:
verified with https://github.com/MarSik/kubevirt-ssp-operator/pull/44

PR #30, opened 2019-04-17
manifests: overhaul the layout

fromanirh:
This patch fixes the manifests layout to:
1. provide a tree layout conformant to https://github.com/operator-framework/operator-registry#manifest-format
   and easy to consume for registry images
2. split the cluster all-in-one manifests (used to make integration with
   HCO easier) into a separate subtree.

The final layout is:
`deploy/`    - master source of truth, everything else is created from the manifests found here
`manifests/` - operator-framework compatible layout
`cluster/`   - amalgamation/all-in-one version of manifests found in `deploy/`, easier to consume (e.g.: HCO integration) - but no new content


fromanirh:
we should wait for the testing pipeline to be set before to merge this. Will require a HCO patch to update the manifest location.

fromanirh:
pasting here comment sent elsewhere, for historical reference:

I think we should merge this.  It's not zero-risk, but it is a great improvement over the current state and will put us in a much better shape should we need more fixes/changes to the CSVs and friends.

PR #20, opened 2019-04-15
Build: rebase on operator-framework images

fromanirh:
Images on quay.io/operator-framework seems to be mantained and
up to date, compared to the old PoC image.

Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #31, opened 2019-05-02
operator: export POD_NAME env var

fromanirh:
Looks like newer version of the ansible operator require this var to be
set. We comply with this patch.

Bug-Url: https://bugzilla.redhat.com/1704914

### DIFF
--- a/cluster/1.0.0/kubevirt-ssp-operator.yaml
+++ b/cluster/1.0.0/kubevirt-ssp-operator.yaml
@@ -213,11 +213,13 @@ spec:
             name: metrics
           imagePullPolicy: Always
           env:
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
             - name: WATCH_NAMESPACE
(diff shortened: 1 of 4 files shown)


tiraboschi:
According to https://github.com/operator-framework/operator-sdk/blob/master/CHANGELOG.md :

> Ansible-based operators have leader election turned on by default. When upgrading, add environment variable POD_NAME to your operator's Deployment using the Kubernetes downward API. To see an example, run operator-sdk new --type=ansible ... and see file deploy/operator.yaml.

since version 0.4 of the operator-sdk.

I'd suggest to start from quay.io/operator-framework/ansible-operator and use the operator-sdk scorecard tool to validate it.

fromanirh:
Thanks @tiraboschi . It seems that https://github.com/MarSik/kubevirt-ssp-operator/pull/20 would have helped catching this early. Will prioritize it.

fromanirh:
tested on local cluster - but in general adding an env var hardly hurts

PR #36, opened 2019-05-13
pkg: update templates version

fromanirh:
The default version we should ship is v0.6.0


fromanirh:
/cc @MarSik @ksimon1 
this value is consumed by HCO. We were deploying an old version, this is surely wrong. Most likely not the only bug that caused https://bugzilla.redhat.com/show_bug.cgi?id=1709268

We will need a HCO PR, will do once we fully figured out what's wrong

PR #37, opened 2019-05-14
validator: prefer deploy in the CR namespace

fromanirh:
Install the validator components into the namespace of the detected CR instead of using the fixed namespace name.
This fixes integration with HCO.


fromanirh:
@MarSik @yanirq @ksimon1 please review, I'm not 100% confident with the ansible filtering yet.

PR #38, opened 2019-05-14
Manifests 1.0.1

fromanirh:


fromanirh:
Verified:
- deployed the operator in non-kubevirt-ns (kubevirt-hyperconverged to mimic the HCO flow
- fixed permissions to let the operator container run (we can assume HCO does this for us :))
- sent validator CR with namespace matching the operator, in general with explicit namespace
- validator succesfully created

cluster/1.0.1/kubevirt-ssp-operator-cr.yaml
```diff
@@ -0,0 +1,22 @@
+---
+apiVersion: kubevirt.io/v1
+kind: KubevirtCommonTemplatesBundle
+metadata:
+  name: kubevirt-common-template-bundle
+spec:
+  version: v0.4.1
```
MarSik: Hmm, how come the version here was not bumped to 0.6.0?

fromanirh: right, fixed

PR #39, opened 2019-05-15
validator: implement support for "automatic" version

fromanirh:
Support a special magic value to let the operator decide
the best version to pull.

The alternative is to just leave the "version" field blank
and leverage the defaults we currently have.

### DIFF
--- a/roles/KubevirtTemplateValidator/templates/service.yaml.j2
+++ b/roles/KubevirtTemplateValidator/templates/service.yaml.j2
@@ -58,7 +58,8 @@ spec:
       serviceAccountName: template-validator
       containers:
         - name: webhook
-          image: {{ ssp_registry | default("quay.io/fromani") }}/{{ validator_image }}:{{ version }}
+          # `version` comes from the CR, while `validator_version` is our computed default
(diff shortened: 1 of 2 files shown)


fromanirh:
@MarSik let's evaluate this approach

roles/KubevirtTemplateValidator/templates/service.yaml.j2
```diff
@@ -58,7 +58,8 @@ spec:
       serviceAccountName: template-validator
       containers:
         - name: webhook
-          image: {{ ssp_registry | default("quay.io/fromani") }}/{{ validator_image }}:{{ version }}
+          # `version` comes from the CR, while `validator_version` is our computed default
+          image: {{ ssp_registry | default("quay.io/fromani") }}/{{ validator_image }}:{{ (version == "automatic") | ternary(validator_version, version) }}
```
MarSik: We should handle the case where no version is provided by the CR or make the version field mandatory in the CRD, right?

fromanirh:
turns out we don't need this in the short term

PR #40, opened 2019-05-15
labeller: manifests: no need for versions in file names

fromanirh:
So we can safely support versionless CRs, letting
the operator decide which version is the best one.

This also cleanups the template lookup paths.


roles/KubevirtNodeLabeller/tasks/main.yml
```diff
@@ -2,13 +2,13 @@
 - name: Create the node labeller roles
   k8s:
     state: present
-    definition: "{{ lookup('template', '/opt/ansible/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-roles-' + version + '.yaml') }}"
+    definition: "{{ lookup('template', '/opt/ansible/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-roles.yaml.j2') }}"
```
fromanirh: I don't get why we use abspaths here, but material for another, future PR

ksimon1: when i wrote this code, i was ansible newbie, so i wanted to be sure, it finds correct files :) Now i would write it differently

MarSik: I just hope the runner is the same upstream and downstream :) I would actually change it right here as it is not worth it to create another patch.

fromanirh: > when i wrote this code, i was ansible newbie, so i wanted to be sure, it finds correct files :) Now i would write it differently

That's fine! The only lesson is just record our decisions in the commit messages and/or comments for future reference.

PR #43, opened 2019-05-20
operator: watch all the cluster namespaces

fromanirh:
And not just the namespace the operator is running into.
We alread y adding ClusterRole{,Binding}s, so permissions should be
fine.

Please note: the automation _dropped_ 'value: ""' from the
new WATCH_NAMESPACE variable, had to reintroduce it manually

Reference: https://github.com/operator-framework/operator-sdk/blob/master/doc/operator-scope.md


fromanirh:
@MarSik I think this change deserves a new tag (1.0.2)

PR #44, opened 2019-05-22
functests: bootstrap and initial batch of tests

fromanirh:
We start testing the deployment of the common templates.
This patch also paves the road for future tests, e.g. about the
validator and the labeller.
We include an initial trivial test for the validator as reference.

The functests assume a running OCP/OKD cluster, see README.md

Running time:
```bash
* [01-test-operator-running.sh                                     ] OK
* [02-test-deploy-templates-openshift.sh                           ] OK
* [03-test-deploy-templates-different-namespace.sh                 ] OK
* [04-test-deploy-templates-operator-namespace.sh                  ] OK
* [05-test-deploy-template-validator.sh                            ] OK
real 167.37
user 14.94
sys 2.67
```

### DIFF
--- a/functests/06-test-deploy-node-labeller.sh
+++ b/functests/06-test-deploy-node-labeller.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+SCRIPTPATH=$( dirname $(readlink -f $0) )
+source ${SCRIPTPATH}/testlib.sh
+
+[ -z ${KV_NAMESPACE} ] && exit 127
+
+RET=1
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/node-labeller-unversioned-cr.yaml" || exit 2
+# TODO: SSP-operator needs to improve its feedback mechanism
+sleep 10s
+for idx in $( seq 1 30); do
+	ANNOTATIONS=$( oc get nodes -o json | jq '.items[0].metadata.annotations | keys' )
+	# any random CPU model annotation that *must* be present if everything's OK
+	HAS_CPU=$( echo ${ANNOTATIONS} | jq 'map(endswith("cpu-model-kvm64")) | any' )
+	# any random KVM info annotation that *must* be present if everything's OK
+	HAS_KVM=$( echo ${ANNOTATIONS} | jq 'map(endswith("kvm-info-cap-hyperv-base")) | any' )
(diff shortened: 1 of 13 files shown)


fromanirh:
/unhold, reviewable now, I consider this the first step and ready to be reviewed.
```bash
# time -p ./test-runner.sh 
* [01-test-operator-running.sh                                     ] OK
* [02-test-deploy-templates-openshift.sh                           ] OK
* [03-test-deploy-templates-different-namespace.sh                 ] OK
* [04-test-deploy-templates-operator-namespace.sh                  ] OK
* [05-test-deploy-template-validator.sh                            ] OK
* [06-test-deploy-node-labeller.sh                                 ] OK
real 152.79
user 14.50
sys 2.57
```

functests/06-test-deploy-node-labeller.sh
```diff
@@ -0,0 +1,28 @@
...
+
+[ -z ${KV_NAMESPACE} ] && exit 127
+
+RET=1
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/node-labeller-unversioned-cr.yaml" || exit 2
+# TODO: SSP-operator needs to improve its feedback mechanism
+sleep 10s
+for idx in $( seq 1 30); do
```
ksimon1: I am not sure if 70 seconds is enough for node labeller, sometimes it is really slow to download all images. Maybe i would do something like this: https://github.com/kubevirt/node-labeller/pull/12/files#diff-80e9ee6c88bb79025632ec61e99d4feeR18
and add there some timeout counter.

fromanirh: Thanks for the heads up. My first thought is we should try to solve this from a different perspective, pre-downloading all the needed images. I'll try this direction first.

fromanirh: While I work on the approch I thought, I added code to implement Karel's suggestion.

MarSik:
I like it, I just wonder if we should use bash or pytest + openshift rest client library as cnv-tests do. I have no strong preference though.

fromanirh:
> I like it, I just wonder if we should use bash or pytest + openshift rest client library as cnv-tests do. I have no strong preference though.

I picked bash only because it is (or it seemed so :)) the quickest way forward.
I have no objection moving to pytest, but I do want to have a testsuite we can use verbatim u/s and d/s.

fromanirh:
Let's start like that and evolve later.

PR #46, opened 2019-05-23
playbooks: enable the repoinfo role

fromanirh:
We should switch this on now, time has come.


fromanirh:
works, verified with https://github.com/MarSik/kubevirt-ssp-operator/pull/44 - but we CANNOT merge this until kubevirt-template-validator is moved under the kubevirt organization

PR #49, opened 2019-05-28
roles: ansible 2.6 compatibility

fromanirh:
The 'k8s_facts' module was introduced in ansible 2.7 (see
https://docs.ansible.com/ansible/latest/modules/k8s_facts_module.html)
but the code needs to be ansible 2.6 - compatible

### DIFF
--- a/roles/KubevirtNodeLabeller/tasks/main.yml
+++ b/roles/KubevirtNodeLabeller/tasks/main.yml
@@ -12,15 +12,9 @@
   register: nl
 
 - name: "Wait for the node-labeller to start"
-  k8s_facts:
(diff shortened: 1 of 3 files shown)


MarSik:
Why 2.6 only? The current Ansible version is 2.8 and we control the environment.

fromanirh:
> Why 2.6 only? The current Ansible version is 2.8 and we control the environment.

Problem, is this turned out to be untrue, at least for a critical corner case. We need to be 2.6 compatible for the time being.

MarSik:
Ack. Using lookup is totally fine, it might even be better semantically as we are executing everything from the controller node anyway.

PR #54, opened 2019-05-30
virt-launcher image URL is now fetched from enviroment variable

ksimon1:
This PR adds ability to set virt-launcher image URL via enviroment variable. 

HCO can now set version of virt-launcher which is used by node-labeller via container enviroment variable.
Thanks to that, node labeller can use the same version of virt-launcher as other components.
When VIRT_LAUNCHER_IMAGE variable is not set, then default value (currently located in roles/KubevirtNodeLabeller/defaults/main.yml) is set.

### DIFF
--- a/deploy/olm-catalog/kubevirt-ssp-operator/1.0.3/kubevirt-ssp-operator.v1.0.3.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/kubevirt-ssp-operator/1.0.3/kubevirt-ssp-operator.v1.0.3.clusterserviceversion.yaml
@@ -1,0 +1,164 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[{"apiVersion":"kubevirt.io/v1","kind":"KubevirtCommonTemplatesBundle","metadata":{"name":"kubevirt-common-template-bundle"},"spec":{"version":"v0.6.0"}},{"apiVersion":"kubevirt.io/v1","kind":"KubevirtNodeLabellerBundle","metadata":{"name":"kubevirt-node-labeller-bundle"},"spec":{"version":"v0.0.5"}},{"apiVersion":"kubevirt.io/v1","kind":"KubevirtTemplateValidator","metadata":{"name":"kubevirt-template-validator","namespace":"kubevirt"},"spec":{"version":"v0.4.8"}}]'
+    capabilities: Basic Install
+  name: kubevirt-ssp-operator.v1.0.3
+  namespace: placeholder
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - kind: KubevirtCommonTemplatesBundle
+      name: kubevirtcommontemplatesbundles.kubevirt.io
+      version: v1
+    - kind: KubevirtNodeLabellerBundle
+      name: kubevirtnodelabellerbundles.kubevirt.io
+      version: v1
+    - kind: KubevirtTemplateValidator
+      name: kubevirttemplatevalidators.kubevirt.io
+      version: v1
+  description: Placeholder description
+  displayName: Kubevirt Ssp Operator
+  install:
+    spec:
+      clusterPermissions:
+      - rules:
+        - apiGroups:
+          - oauth.openshift.io
+          - template.openshift.io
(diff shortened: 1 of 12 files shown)


ksimon1:
cc @fromanirh, @tiraboschi

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
```diff
@@ -44,7 +44,7 @@ spec:
               name: cpu-config
 
         - name: libvirt
-          image: {{ ssp_registry | default('kubevirt') }}/{{ libvirt_image }}:{{ libvirt_tag }}
+          image: {{ lookup('env','VIRT_LAUNCHER_IMAGE')| default(virt_launcher_image, true) }}
```
fromanirh: please elaborate about what this line is supposed to do should VIRT_LAUNCHER_IMAGE not be defined

fromanirh: in retrospect I think it would be better to use two env vars, one for the libvirt_image and one for the libvirt_tag. I think it should look like

 image: {{ ssp_registry | default('kubevirt') }}/{{ lookup('env','VIRT_LAUNCHER_IMAGE')| default(libvirt_image, true) }}:{{ lookup('env','VIRT_LAUNCHER_TAG')| default(libvirt_tag, true) }}

fromanirh: in retrospect I think it would be better to use two env vars, one for the libvirt_image and one for the libvirt_tag. I think it should look like
```
 image: {{ ssp_registry | default('kubevirt') }}/{{ lookup('env','VIRT_LAUNCHER_IMAGE')| default(libvirt_image, true) }}:{{ lookup('env','VIRT_LAUNCHER_TAG')| default(libvirt_tag, true) }}
```

ksimon1: @tiraboschi can the script which builds CSV replace only e.g. part of value?

tiraboschi: Now it's replacing only the TAG

fromanirh: Replacing the tag would actually address my concern above! we can keep the ssp_registry variable and get the right tag value from HCO.
@ksimon1 , do you think you can update the PR to make this happen?

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
```diff
@@ -44,7 +44,7 @@ spec:
               name: cpu-config
 
         - name: libvirt
-          image: {{ ssp_registry | default('kubevirt') }}/{{ libvirt_image }}:{{ libvirt_tag }}
+          image: {{ ssp_registry | default('kubevirt') }}/{{ virt_launcher_image }}:{{ lookup('env','VIRT_LAUNCHER_TAG')| default(virt_launcher_tag, true) }}
```
fromanirh: LGTM

deploy/olm-catalog/kubevirt-ssp-operator/1.0.3/kubevirt-ssp-operator.v1.0.3.clusterserviceversion.yaml
```diff
@@ -0,0 +1,164 @@
...
+            spec:
+              containers:
+              - env:
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: WATCH_NAMESPACE
+                - name: VIRT_LAUNCHER_TAG
+                  value: v0.17.0
```
fromanirh: @tiraboschi this is how it is supposed to look, right?

tiraboschi: yes, right

fromanirh:
Thanks @tiraboschi !

PR #56, opened 2019-05-31
Traviskube: add travis integration

fromanirh:
This PR make travis run functests, courtesy of traviskube (github.com/fabiand/traviskube)
I actually forked traviskube to add some SSP-specific features.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "ci"]
+	path = ci
+	url = https://github.com/fromanirh/traviskube
```
yanirq: What happens here if we move to kubevirt offical repo ?

fromanirh: for this specific hunk I suggest to keep it like this to point to my repo for the time being. I have few changes that I'm not sure they can make quickly into u/s - and I surely need to polish them a bit.
After that we will have to change the git submodule ref.

hack/release-manifests.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -ex
+
+SELF=$( realpath $0 )
+BASEPATH=$( dirname $SELF )
+
+TAG="${1:-v0.0.0}"
+VERSION=$( ${BASEPATH}/../build/version-from-tag.py ${TAG} )
```
MarSik: Why not pure bash? ${TAG#v} removes prefix "v" from variable TAG

fromanirh: Only because I forgot the syntax :) Fixing

Makefile
```diff
@@ -20,3 +21,21 @@ container-push-operator:
 
 container-push-registry:
 	docker push $(IMAGE_REGISTRY)/$(REGISTRY_IMAGE):$(IMAGE_TAG)
+
+operator-sdk:
+	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
+	chmod 0755 operator-sdk
+
+_out:
```
yanirq: not sure if its critical here, but maybe delete the directory if exists and re create it (if there is a possible scenario of existing artefacts)

fromanirh: Probably, not critical, but won't hurt. Will fix.

yanirq:
If think the title for this PR can be more general : add travis integration.   you are not only functests but also an automated release process

fromanirh:
@yanirq good point. PR renamed.

.travis.yml
```diff
@@ -17,6 +17,25 @@ jobs:
...
+      skip_cleanup: true
+      overwrite: true
+      file_glob: true
+      file:
+      - _out/*
+      name: $TRAVIS_TAG
+      on:
+        tags: true
+        branch: master
+        repo: MarSik/kubevirt-ssp-operator
```
MarSik: Looks like we could use $TRAVIS_REPO_SLUG to get this info. It might help when we decide to move under kubevirt.

fromanirh:
Unfortunately there is not yet a clear logic or process to choose the timeouts/sleeps, nor a better sync mechanism is available.
All magic numbers have been chosed after some trial/error, and they are subject to change. There are not "right" numbers yet.

fromanirh:
We agreed I'll post a followup PR with smarter approach to timeouts, e.g. using `oc wait`. This one is already too big.

fromanirh:
after few more fixes, looks good https://travis-ci.org/fromanirh/kubevirt-ssp-operator/builds/541700424

yanirq:
lgtm

PR #57, opened 2019-06-06
docs: install reuse the travis script

fromanirh:
With minor changes, we can provide a simple install
script for everyone brave enough to install the
SSP-operator in their cluster.

This simplifies the install process quite a lot
with no real drawbacks.

### DIFF
--- a/hack/install-operator.sh
+++ b/hack/install-operator.sh
@@ -6,6 +6,16 @@ BASEPATH=$( dirname $SELF )
 
 NAMESPACE=${1:-kubevirt}
 
+_curl() {
+	# this dupes the baseline "curl" command line, but is simpler
+	# wrt shell quoting/expansion.
+	if [ -n "${GITHUB_TOKEN}" ]; then
+		curl -H "Authorization: token ${GITHUB_TOKEN}" $@
+	else
+		curl $@
+	fi
+}
+
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_commontemplatesbundle_crd.yaml
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_nodelabellerbundle_crd.yaml
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_templatevalidator_crd.yaml
@@ -13,7 +23,7 @@ oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_templatevalidator_crd.yaml
 LAST_TAG=""
 if [ "${CI}" != "true" ] || [ "${TRAVIS}" != "true" ]; then
 	# TODO: consume releases, not tags
-	LAST_TAG=$( curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[0].name' )
+	LAST_TAG=$( _curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[0].name' )
 fi
 
 oc create -n ${NAMESPACE} -f ${BASEPATH}/../deploy/service_account.yaml
--- a/hack/install-operator.sh
+++ b/hack/install-operator.sh
@@ -23,7 +23,8 @@ oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_templatevalidator_crd.yaml
 LAST_TAG=""
 if [ "${CI}" != "true" ] || [ "${TRAVIS}" != "true" ]; then
 	# TODO: consume releases, not tags
-	LAST_TAG=$( _curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[0].name' )
+	# TODO: check if the github APIs guarantee the ordering
+	LAST_TAG=$( _curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[].name' | sort -r | head -1 )
 fi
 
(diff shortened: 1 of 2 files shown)


fromanirh:
@MarSik PR failed CI because of github throttling :\

alexxa:
Works for me
```
(env) # hack/install-operator.sh 
serviceaccount/kubevirt-ssp-operator created
clusterrole.rbac.authorization.k8s.io/cluster-admin added: "system:serviceaccount:kubevirt:kubevirt-ssp-operator"
clusterrole.rbac.authorization.k8s.io/kubevirt-ssp-operator created
clusterrolebinding.rbac.authorization.k8s.io/kubevirt-ssp-operator created
customresourcedefinition.apiextensions.k8s.io/kubevirtcommontemplatesbundles.kubevirt.io created
customresourcedefinition.apiextensions.k8s.io/kubevirtnodelabellerbundles.kubevirt.io created
customresourcedefinition.apiextensions.k8s.io/kubevirttemplatevalidators.kubevirt.io created
deployment.apps/kubevirt-ssp-operator created
(env) # oc get pods -n kubevirt
NAME READY STATUS RESTARTS AGE
kubevirt-ssp-operator-c4467b768-cpsvm 0/1 ContainerCreating 0 10s
(env) # oc get pods -n kubevirt
NAME READY STATUS RESTARTS AGE
kubevirt-ssp-operator-c4467b768-cpsvm 1/1 Running 0 48s
```

MarSik:
@fromanirh This looks like we need to pass the token to traviscube somehow.

fromanirh:
> @fromanirh This looks like we need to pass the token to traviscube somehow.

I set the GITHUB_TOKEN env var with a personal access token (https://github.com/settings/tokens) in my own fork of this repo, perhaps we can do the same here?

hack/install-operator.sh
```diff
@@ -6,7 +6,11 @@ BASEPATH=$( dirname $SELF )
 
 NAMESPACE=${1:-kubevirt}
 
-REGISTRY=$(minishift openshift registry)
+LAST_TAG=""
+if [ "${CI}" != "true" ] || [ "${TRAVIS}" != "true" ]; then
+	# TODO: consume releases, not tags
+	LAST_TAG=$( curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[0].name' )
```
MarSik: Note you are using simple curl here as well. This will be affected by the api throttling as the environment variable is not automatically passed in. You need to send a token explicitely: https://developer.github.com/v3/auth/

MarSik:
We need to see if traviscube knows how to consume the token. Curl will not pass it to github it by default.

hack/install-operator.sh
MarSik: On the other hand, this specific line is not going to be called from CI. But developers from the same office might still hit this (too many people with the same IP)

fromanirh: @MarSik ack. instructions are indeed simple (https://developer.github.com/v3/auth/#via-oauth-tokens), but I wonder which pair user/token should i use. Mine? Yours? something else?

MarSik: I would use the GITHUB_TOKEN environment variable. We can then generate a proper token and add it to Travis.

Or GITHUB_TOKEN_API with fallback to GITHUB_TOKEN to allow for separate release (with permissions) and read only API (no extra permissions) tokens.

fromanirh:
> We need to see if traviscube knows how to consume the token. Curl will not pass it to github it by default.

PR for my fork of traviskube here: https://github.com/fromanirh/traviskube/pull/1

hack/install-operator.sh
fromanirh: OK, so https://github.com/fromanirh/traviskube/pull/1 is a good direction :) Let's continue here the discussion, then I'll update this PR accordingly

fromanirh: Fixed in my traviskube fork.

hack/install-operator.sh
```diff
@@ -6,24 +6,46 @@ BASEPATH=$( dirname $SELF )
...
+}
+
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_commontemplatesbundle_crd.yaml
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_nodelabellerbundle_crd.yaml
 oc create -f ${BASEPATH}/../deploy/crds/kubevirt_v1_templatevalidator_crd.yaml
 
+LAST_TAG=""
+if [ "${CI}" != "true" ] || [ "${TRAVIS}" != "true" ]; then
+	# TODO: consume releases, not tags
+	LAST_TAG=$( _curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/tags | jq -r '.[0].name' )
```
ksimon1: are these releases sorted from newest to oldest? Because e.g. you create release 1.2.0 and then you realize in 1.1.0 was mistake, so you create 1.1.1 release and now newest release is 1.1.1 not 1.2.0

fromanirh: Good point. I should check and make sure to pick the latest.

ksimon1: This might solve it:
`curl -s https://api.github.com/repos/MarSik/kubevirt-ssp-operator/releases|grep tag_name|sort -V | tail -1 | awk -F':' '{print $2}' | sed 's/,//' | xargs`
(taken from https://kubevirt.io/quickstart_minikube/)

PR #58, opened 2019-06-11
manifests: install: fix RBAC permissions

fromanirh:
The PR https://github.com/MarSik/kubevirt-ssp-operator/pull/27
 had a bug which was not evident on standalone installation, because in
the install script we granted too broad permissions.

This became evident when deployed under HCO.
This PR aims to fix the RBAC permission issues without using too broad
permission sets.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -4,6 +4,7 @@ metadata:
   name: kubevirt-ssp-operator
 rules:
 - apiGroups:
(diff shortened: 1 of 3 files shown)


fromanirh:
also, verified running the template-validator functests on a validator deployed by patched SSP operator;
also verified templates are deployed and labeller. node labels looks good.

PR #59, opened 2019-06-13
version of ssp components can be now set via operator enviroment variable

ksimon1:
This PR enables to set version of ssp components via container enviroment variables.

Updated node labeller, validator to new version

### DIFF
--- a/_defaults.yml
+++ b/_defaults.yml
@@ -11,3 +11,11 @@
 # validator_tag: v0.4.8
 # templates_version: v0.6.0
 ---
+#node-labeller
+kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.0', true) }}"
+kvm_info_nfd_plugin_tag: "{{ lookup('env','KVM_INFO_TAG')| default('v0.5.7', true) }}"
+kubevirt_cpu_nfd_plugin_tag: "{{ lookup('env','CPU_PLUGIN_TAG')| default('v0.1.0', true) }}"
+virt_launcher_tag: "{{ lookup('env','VIRT_LAUNCHER_TAG')| default('v0.18.0', true) }}"
+
+#validator
+validator_tag: "{{ lookup('env','VALIDATOR_TAG')| default('v0.6.1', true) }}"
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -28,7 +28,15 @@ spec:
                   fieldPath: metadata.name
             - name: WATCH_NAMESPACE
               value: ""
+            - name: KVM_INFO_TAG
+              value: ""
+            - name: VALIDATOR_TAG
+              value: ""
             - name: VIRT_LAUNCHER_TAG
-              value: "v0.17.0"
+              value: ""
+            - name: NODE_LABELLER_TAG
+              value: ""
+            - name: CPU_PLUGIN_TAG
+              value: ""
             - name: OPERATOR_NAME
               value: "kubevirt-ssp-operator"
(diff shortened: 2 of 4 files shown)


ksimon1:
/cc @MarSik, @fromanirh 
Do you know, if there are any other places, which has to be changed?

fromanirh:
@ksimon1 your approach works, but can we just move all the logic that fetches the env vars and fallbacks to the hardcoded defaults in _defaults.yml?
I think that was the original intent

ksimon1:
@fromanirh i updated the code with latest changes.

deploy/operator.yaml
```diff
@@ -28,7 +28,17 @@ spec:
                   fieldPath: metadata.name
             - name: WATCH_NAMESPACE
               value: ""
+            - name: TEMPLATES_TAG
```
MarSik: I would actually leave the manifest intact and provide the fallback values in _defaults.yaml. That way we do not have to change that many files every time a version changes.

ksimon1: But if we remove the variables from operator.yaml, then HCO will not be able to set correct values. (or you meant it like keep variables here and just remove values?)

ksimon1:
@tiraboschi please can you review this?

_defaults.yml
```diff
@@ -11,3 +11,11 @@
...
 # templates_version: v0.6.0
 ---
+#node-labeller
+kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.0', true) }}"
+kvm_info_nfd_plugin_tag: "{{ lookup('env','KVM_INFO_TAG')| default('v0.5.7', true) }}"
+kubevirt_cpu_nfd_plugin_tag: "{{ lookup('env','CPU_PLUGIN_TAG')| default('v0.1.0', true) }}"
+virt_launcher_tag: "{{ lookup('env','VIRT_LAUNCHER_TAG')| default('v0.18.0', true) }}"
+
+#validator
+validator_tag: "{{ lookup('env','VALIDATOR_TAG')| default('v0.4.8', true) }}"
```
fromanirh: let's use v0.6.1 as hardcoded default please

ksimon1: done

roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -2,11 +2,7 @@
 # defaults file for KubevirtNodeLabeller
 wait: true
 
-kubevirt_node_labeller_image: "kubevirt-node-labeller"
-kubevirt_node_labeller_tag: "v0.0.5"
```
MarSik: I would actually keep tags here as well to make sure it all works even without the _default file. But set them to "latest" so we do not have to maintain the versions here.

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
```diff
@@ -33,9 +33,9 @@ spec:
               mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
 
         - name: kubevirt-cpu-nfd-plugin
-          image: {{ ssp_registry | default('quay.io/ksimon') }}/{{ kubevirt_cpu_nfd_plugin_image }}:{{ kubevirt_cpu_nfd_plugin_tag }}
+          image: {{ ssp_registry | default('quay.io/kubevirt') }}/{{ kubevirt_cpu_nfd_plugin_image }}:{{ kubevirt_cpu_nfd_plugin_tag }}
```
MarSik: Since the registry is now the same for everything, what about setting ssp_registry in the role defaults yaml and keeping just `{{ ssp_registry }}` in the code?

ksimon1: virt-launcher image uses different registry (dockerhub).

MarSik: I see. Well then we have to keep it like it is.

PR #62, opened 2019-07-04
reduce operator permissions

ksimon1:
This PR reduces ssp operator permission

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -5,23 +5,25 @@ metadata:
 rules:
 - apiGroups:
   - kubevirt.io
-  - oauth.openshift.io
   - template.openshift.io
   resources:
   - '*'
   verbs:
-  - '*'
+  - create 
+  - get
+  - list
+  - patch
+  - update
+  - watch
+
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
   - clusterroles
   verbs:
   - create
-  - get
-  - list
   - watch
   - patch
-  - update
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
@@ -32,7 +34,6 @@ rules:
   - list
   - watch
   - patch
-  - update
 - apiGroups:
   - extensions
   - apps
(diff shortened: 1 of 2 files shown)


deploy/role.yaml
```diff
@@ -5,23 +5,25 @@ metadata:
...
   - '*'
   verbs:
-  - '*'
+  - create 
+  - get
+  - list
+  - patch
+  - update
+  - watch
+
```
fromanirh: For (my) information: isn't this the whole list implied by '*'? Better explicit than implicit, though.

deploy/role.yaml
```diff
@@ -5,23 +5,25 @@ metadata:
...
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
   - clusterroles
   verbs:
   - create
-  - get
-  - list
   - watch
   - patch
```
fromanirh: do we also need "patch"?

deploy/role.yaml
```diff
@@ -32,49 +34,45 @@ rules:
...
-  - update
 - apiGroups:
   - extensions
   - apps
   resources:
   - deployments
   - replicasets
   - daemonsets
-  - statefulsets
```
fromanirh: this was added because older k8s had a quirk that required permission for daemonsets *and* statefulsets to work properly. Probably it's indeed time to remove it.

deploy/role.yaml
```diff
@@ -32,49 +34,45 @@ rules:
...
+  - watch
 - apiGroups:
   - ""
   resources:
   - serviceaccounts
   - configmaps
-  - nodes
-  - endpoints
-  - events
-  - secrets
```
fromanirh: yep, unneeded because we don't create our own secrets (for template-validator) anymore

deploy/role.yaml
```diff
@@ -32,49 +34,45 @@ rules:
...
+  - get
+  - patch
+  - update
 - apiGroups:
   - ""
   resources:
-  - services
+  - pods
   verbs:
-  - create
```
fromanirh: interesting, so dont' we need to create/list/watch pods?

deploy/role.yaml
ksimon1: '*' should contain three more actions: delete, deletecollection, proxy (i am not sure about this one)

deploy/role.yaml
ksimon1: yep we need it:
`system:serviceaccount:kubevirt:kubevirt-ssp-operator" cannot patch clusterroles.rbac.authorization.k8s.io at the cluster scope: no RBAC policy matched", "reason": "Forbidden"`

deploy/role.yaml
ksimon1: SSP operator doesn't complain, that it doesn't have these permissions.

deploy/role.yaml
fromanirh: ok

deploy/role.yaml
fromanirh: fair enough

ksimon1:
@MarSik can we merge this PR? or anything else has to be done?

PR #64, opened 2019-07-08
build: don't include py source code in the image

fromanirh:
Let's have a binary-only top-level image.
Assuming that ansible playbooks don't count as source code :)


MarSik:
Hmm, is there any benefit in making binary only image? Maybe size? This might also make debugging harder if we ever decide to ship ansible plugins.

fromanirh:
obsolete

PR #66, opened 2019-07-29
Add csv-generator logic to ssp operator container image

davidvossel:
This PR adds the HCO's `csv-generator` logic to the ssp operator's container image. 

## Overview of Changes
* New `make csv-generator` make target generates a reusable pre-generated CSV file with placeholder values in the manifests/generated dir.
* `make container-build-operator` copies in both the placeholder CSV file and the new `csv-generator` script
* The operator's container can now be used to generate a corresponding CSV file using the `csv-generator` entry-point.

**Basic Example**
`docker run -it -rm --entrypoint=/usr/bin/csv-generator <operator image> --csv-version=1.1.1 --operator-image=someregistry/ssp-opereator:sometag --namespace=hco-namespace`

The HCO will be using this csv-generator entrypoint in the process of creating the aggregated HCO CSV that encompasses multiple component-level operators (like ssp).

### DIFF
--- a/build/csv-generator.sh
+++ b/build/csv-generator.sh
@@ -31,6 +31,7 @@ help_text() {
     echo "  --virt-launcher-tag: (OPTIONAL)"
     echo "  --node-labeller-tag: (OPTIONAL)"
     echo "  --cpu-plugin-tag:    (OPTIONAL)"
+    echo "  --dump-crds:         (OPTIONAL) Dumps CRD manifests with the CSV to stdout"
 }
 
 # REQUIRED ARGS
@@ -45,6 +46,7 @@ VALIDATOR_TAG=""
 VIRT_LAUNCHER_TAG=""
 NODE_LABELLER_TAG=""
 CPU_PLUGIN_TAG=""
+DUMP_CRDS=""
 
 while (( "$#" )); do
     ARG=`echo $1 | awk -F= '{print $1}'`
@@ -79,6 +81,9 @@ while (( "$#" )); do
     --cpu-plugin-tag)
         CPU_PLUGIN_TAG=$VAL
         ;;
+    --dump-crds)
+        DUMP_CRDS="true"
+        ;;
     --)
         break
         ;;
@@ -100,8 +105,8 @@ cp ${MANIFESTS_GENERATED_CSV} ${TMP_FILE}
 # replace placeholder version with a human readable variable name
 # that will be used later on by csv-generator
 sed -i "s/PLACEHOLDER_CSV_VERSION/${CSV_VERSION}/g" ${TMP_FILE}
-sed -i "s/namespace: placeholder/${NAMESPACE}/g" ${TMP_FILE}
-sed -i "s/REPLACE_IMAGE/${OPERATOR_IMAGE}/g" ${TMP_FILE}
+sed -i "s/namespace: placeholder/namespace: ${NAMESPACE}/g" ${TMP_FILE}
+sed -i "s|REPLACE_IMAGE|${OPERATOR_IMAGE}|g" ${TMP_FILE}
 
 replace_env_var "WATCH_NAMESPACE" $WATCH_NAMESPACE
 replace_env_var "KVM_INFO_TAG" $KVM_INFO_TAG
@@ -113,8 +118,10 @@ replace_env_var "CPU_PLUGIN_TAG" $CPU_PLUGIN_TAG
 # dump CSV and CRD manifests to stdout
 echo "---"
 cat ${TMP_FILE}
-for CRD in $( ls deploy/crds/kubevirt_*crd.yaml ); do
-	echo "---"
-	cat ${CRD}
-done
 rm ${TMP_FILE}
+if [ "$DUMP_CRDS" = "true" ]; then
+    for CRD in $( ls ${CRDS_DIR}/kubevirt_*crd.yaml ); do
+        echo "---"
+        cat ${CRD}
+    done
+fi
--- a/build/csv-generator.sh
+++ b/build/csv-generator.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+set -e
+
+MANIFESTS_GENERATED_DIR="manifests/generated"
+if ! [ -d $MANIFESTS_GENERATED_DIR ]; then
+    MANIFESTS_GENERATED_DIR="${HOME}/manifests/generated"
+fi
+MANIFESTS_GENERATED_CSV=${MANIFESTS_GENERATED_DIR}/kubevirt-ssp-operator.vVERSION.clusterserviceversion.yaml
+TMP_FILE=$(mktemp)
+
+replace_env_var() {
+    local value_offset="                  "
+    local key=$1
+    local var=$2
+    sed -i "s/- name: ${key}/- name: $1\n${value_offset}value: \"${var}\"/g" ${TMP_FILE}
+}
+
+help_text() {
+    echo "USAGE: csv-generator --csv-version=<version> --namespace=<namespace> --operator-image=<operator image> [optional args]"
+    echo ""
+    echo "ARGS:"
+    echo "  --csv-version:    (REQUIRED) The version of the CSV file"
+    echo "  --namespace:      (REQUIRED) The namespace set on the CSV file"
+    echo "  --operator-image: (REQUIRED) The operator container image to use in the CSV file"
+    echo "  --watch-namespace:   (OPTIONAL)"
+    echo "  --kvm-info-tag:      (OPTIONAL)"
+    echo "  --validator-tag:     (OPTIONAL)"
+    echo "  --virt-launcher-tag: (OPTIONAL)"
+    echo "  --node-labeller-tag: (OPTIONAL)"
+    echo "  --cpu-plugin-tag:    (OPTIONAL)"
+}
+
+# REQUIRED ARGS
+CSV_VERSION=""
+NAMESPACE=""
+OPERATOR_IMAGE=""
+
+# OPTIONAL ARGS
+WATCH_NAMESPACE=""
+KVM_INFO_TAG=""
+VALIDATOR_TAG=""
+VIRT_LAUNCHER_TAG=""
+NODE_LABELLER_TAG=""
+CPU_PLUGIN_TAG=""
+
+while (( "$#" )); do
+    ARG=`echo $1 | awk -F= '{print $1}'`
+    VAL=`echo $1 | awk -F= '{print $2}'`
+    shift
+
+    case "$ARG" in
+    --csv-version)
+        CSV_VERSION=$VAL
+        ;;
+    --namespace)
+        NAMESPACE=$VAL
+        ;;
+    --operator-image)
+        OPERATOR_IMAGE=$VAL
+        ;;
+    --watch-namespace)
+        WATCH_NAMESPACE=$VAL
+        ;;
+    --kvm-info-tag)
+        KVM_INFO_TAG=$VAL
+        ;;
+    --validator-tag)
+        VALIDATOR_TAG=$VAL
+        ;;
+    --virt-launcher-tag)
+        VIRT_LAUNCHER_TAG=$VAL
+        ;;
+    --node-labeller-tag)
+        NODE_LABELLER_TAG=$VAL
+        ;;
+    --cpu-plugin-tag)
+        CPU_PLUGIN_TAG=$VAL
+        ;;
+    --)
+        break
+        ;;
+    *) # unsupported flag
+        echo "Error: Unsupported flag $ARG" >&2
+        exit 1
+        ;;
+    esac
+done
+
+if [ -z "$CSV_VERSION" ] || [ -z "$NAMESPACE" ] || [ -z "$OPERATOR_IMAGE" ]; then
+    echo "Error: Missing required arguments"
+    help_text
+    exit 1
+fi
+
+cp ${MANIFESTS_GENERATED_CSV} ${TMP_FILE}
+
+# replace placeholder version with a human readable variable name
+# that will be used later on by csv-generator
+sed -i "s/PLACEHOLDER_CSV_VERSION/${CSV_VERSION}/g" ${TMP_FILE}
+sed -i "s/namespace: placeholder/${NAMESPACE}/g" ${TMP_FILE}
+sed -i "s/REPLACE_IMAGE/${OPERATOR_IMAGE}/g" ${TMP_FILE}
+
+replace_env_var "WATCH_NAMESPACE" $WATCH_NAMESPACE
+replace_env_var "KVM_INFO_TAG" $KVM_INFO_TAG
+replace_env_var "VALIDATOR_TAG" $VALIDATOR_TAG
+replace_env_var "VIRT_LAUNCHER_TAG" $VIRT_LAUNCHER_TAG
+replace_env_var "NODE_LABELLER_TAG" $NODE_LABELLER_TAG
+replace_env_var "CPU_PLUGIN_TAG" $CPU_PLUGIN_TAG
+
+cat ${TMP_FILE}
+rm ${TMP_FILE}
--- a/build/csv-generator.sh
+++ b/build/csv-generator.sh
@@ -3,8 +3,10 @@
 set -e
 
 MANIFESTS_GENERATED_DIR="manifests/generated"
+CRDS_DIR="deploy/crds"
 if ! [ -d $MANIFESTS_GENERATED_DIR ]; then
     MANIFESTS_GENERATED_DIR="${HOME}/manifests/generated"
+    CRDS_DIR="${HOME}/deploy/crds"
 fi
 MANIFESTS_GENERATED_CSV=${MANIFESTS_GENERATED_DIR}/kubevirt-ssp-operator.vVERSION.clusterserviceversion.yaml
 TMP_FILE=$(mktemp)
@@ -108,5 +110,11 @@ replace_env_var "VIRT_LAUNCHER_TAG" $VIRT_LAUNCHER_TAG
 replace_env_var "NODE_LABELLER_TAG" $NODE_LABELLER_TAG
 replace_env_var "CPU_PLUGIN_TAG" $CPU_PLUGIN_TAG
 
+# dump CSV and CRD manifests to stdout
+echo "---"
 cat ${TMP_FILE}
+for CRD in $( ls deploy/crds/kubevirt_*crd.yaml ); do
+	echo "---"
+	cat ${CRD}
+done
 rm ${TMP_FILE}
--- a/build/make-csv-generator.sh
+++ b/build/make-csv-generator.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+if [ -x "${BASEPATH}/../operator-sdk" ]; then
+       OPERATOR_SDK="${BASEPATH}/../operator-sdk"
+else       
+    which operator-sdk &> /dev/null || {
+        echo "operator-sdk not found (see https://github.com/operator-framework/operator-sdk)"
+        exit 1
+    }
+    OPERATOR_SDK="operator-sdk"
+fi
+
+MANIFESTS_GENERATED_DIR="manifests/generated"
+MANIFESTS_GENERATED_CSV=${MANIFESTS_GENERATED_DIR}/kubevirt-ssp-operator.vVERSION.clusterserviceversion.yaml
+PLACEHOLDER_CSV_VERSION="9999.9999.9999"
+
+# Create CSV with placeholder version. The version
+# has to be semver compatible in order for the 
+# operator sdk to create it for us. That's why we
+# are using the absurd 9999.9999.9999 version here. 
+${OPERATOR_SDK} olm-catalog gen-csv --csv-version ${PLACEHOLDER_CSV_VERSION}
+
+# Move CSV to generated folder
+mv deploy/olm-catalog/kubevirt-ssp-operator/${PLACEHOLDER_CSV_VERSION}/kubevirt-ssp-operator.v${PLACEHOLDER_CSV_VERSION}.clusterserviceversion.yaml $MANIFESTS_GENERATED_CSV
+
+# cleanup placeholder version's deployment dir
+rm -rf mv deploy/olm-catalog/kubevirt-ssp-operator/${PLACEHOLDER_CSV_VERSION}
+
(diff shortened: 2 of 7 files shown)


davidvossel:
/cc @MarSik @irosenzw @fromanirh @rthallisey @djzager

This is the pattern I came up with for the SSP and NMO operators. The NMO PR is still in progress, but it will look similar if we all agree on the approach.

Since these two operators aren't based on GoLang, I opted for a bash script that simply replaces variables in a pre-generated CSV file rather than a full blown GoLang implementation. The part that creates the pre-generated CSV file that `csv-generator` does serach/replaces on is a part of the operator's build step, so it should always be in sync with the operator build.

MarSik:
@davidvossel We already have a script like this - see the file https://github.com/MarSik/kubevirt-ssp-operator/blob/master/build/make-manifests.sh

I am not particularly happy about including this to the image though (everybody needs to download extra data that are not needed). We release the images and manifests using Travis automation and github releases. Is there a reason you can't take it from there? We could add a label with url to the manifest to the image too to make it easy to download.

davidvossel:
> I am not particularly happy about including this to the image though (everybody needs to download extra data that are not needed). We release the images and manifests using Travis automation and github releases. Is there a reason you can't take it from there? We could add a label with url to the manifest to the image too to make it easy to download.

This is for HCO, we need the script in the operator's container image.

yanirq:
>This is for HCO, we need the script in the operator's container image.

I think that if the csv-generator output provides us the same output that `/build/make-manifests.sh` gave us then its a bit of a duplication. 
If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)

davidvossel:
> I think that if the csv-generator output provides us the same output that /build/make-manifests.sh gave us then its a bit of a duplication.
If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)

There is some duplication going on, but it's unclear to me how to converge the two. 

Would it make sense to have csv-generator replace this line in the build/make-manifests.sh file

```${OPERATOR_SDK} olm-catalog gen-csv --csv-version ${VERSION}```

would become something like 

```build/csv-generator --csv-version ${VERSION} --operator-image "REPLACE_IMAGE" --namespace "placeholder"```

?

**Comment by djzager** [comment]
> I am not particularly happy about including this to the image though (everybody needs to download extra data that are not needed).

Are you referring to a shell script that puts a CSV on stdout? Of all the crimes against container image sizes...this doesn't rate objection in my opinion. I don't want to dismiss your concern at the same time I just simply don't agree that it's impact on the container image build size raises the bar for this PR to be accepted.

> I think that if the csv-generator output provides us the same output that `/build/make-manifests.sh` gave us then its a bit of a duplication.
> If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)

I agree that this is a duplication...well, kind of...since one script (which on quick glance doesn't appear to be available on the container image) puts the files on disk (some which are useful and others that are not) while the one proposed here puts the things that the HCO would need on stdout to be consumed in automation.

IMHO discussion of the finer implementation details of this PR, like "why don't we just use the release manifests on github?" or "why not just use `/build/make-manifests.sh`?" are time wasted unless we have an actual plan that either uses the manifest on github or uses `/build/make-manifests.sh` to, in an automated way, build an HCO CSV that is correct.

The primary problem I see with releases is...how do you automate grabbing the correct file from each of the component operators (across multiple releases)... it would be a huge pain if one project is doing their releases one way and another some other way.

The primary problem I see with updating the `build/make-manifests.sh` is that putting the necessary OLM bits on stdout is easier to deal with the files on disk.

**Comment by yanirq** [comment]
> > I think that if the csv-generator output provides us the same output that /build/make-manifests.sh gave us then its a bit of a duplication.
> > If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)
> 
> There is some duplication going on, but it's unclear to me how to converge the two.
> 
> Would it make sense to have csv-generator replace this line in the build/make-manifests.sh file
> 
> `${OPERATOR_SDK} olm-catalog gen-csv --csv-version ${VERSION}`
> 
> would become something like
> 
> `build/csv-generator --csv-version ${VERSION} --operator-image "REPLACE_IMAGE" --namespace "placeholder"`
> 
> ?

Yes, sounds about right.  just cleaning up the duplicates and code.  
Maybe cleanup can be done in a separate PR.

**Comment by davidvossel** [comment]
> Yes, sounds about right for consistency. just cleaning up the duplicates and code.
Maybe cleanup can be done in a separate PR.

I can do it now in this PR.  I'll post an update later today. Since I'll be copying the general approach of whatever we end up going with here for NMO, I think it's worth us sorting out how to handle duplication in order to make this as maintainable as possible.

**Comment by davidvossel** [comment]
> The primary problem I see with updating the build/make-manifests.sh is that putting the necessary OLM bits on stdout is easier to deal with the files on disk.

@djzager good analysis. I think martin's on PTO and simply hasn't gotten a chance to catch up on some of these discussions yet which is why his response doesn't reflect the larger context of what's going on with the HCO and builds. :)

**Comment by fromanirh** [comment]
> > > I think that if the csv-generator output provides us the same output that /build/make-manifests.sh gave us then its a bit of a duplication.
> > > If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)
> > 
> > 
> > There is some duplication going on, but it's unclear to me how to converge the two.
> > Would it make sense to have csv-generator replace this line in the build/make-manifests.sh file
> > `${OPERATOR_SDK} olm-catalog gen-csv --csv-version ${VERSION}`
> > would become something like
> > `build/csv-generator --csv-version ${VERSION} --operator-image "REPLACE_IMAGE" --namespace "placeholder"`
> > ?
> 
> Yes, sounds about right for consistency. just cleaning up the duplicates and code.
> Maybe cleanup can be done in a separate PR.

Totally agree with Yanir. If David can upload a fixed version, even better.

**Comment by fromanirh** [comment]
> > I think that if the csv-generator output provides us the same output that `/build/make-manifests.sh` gave us then its a bit of a duplication.
> > If so, we might just need to clean up and align the makefile and the travis release process that provides us the cvs files according to the version (e.g. when creating a new release tag)
> 
> I agree that this is a duplication...well, kind of...since one script (which on quick glance doesn't appear to be available on the container image) puts the files on disk (some which are useful and others that are not) while the one proposed here puts the things that the HCO would need on stdout to be consumed in automation.
> 
> IMHO discussion of the finer implementation details of this PR, like "why don't we just use the release manifests on github?" or "why not just use `/build/make-manifests.sh`?" are time wasted unless we have an actual plan that either uses the manifest on github or uses `/build/make-manifests.sh` to, in an automated way, build an HCO CSV that is correct.
> 
> The primary problem I see with releases is...how do you automate grabbing the correct file from each of the component operators (across multiple releases)... it would be a huge pain if one project is doing their releases one way and another some other way.
> 
> The primary problem I see with updating the `build/make-manifests.sh` is that putting the necessary OLM bits on stdout is easier to deal with the files on disk.

Hey, I just would like to clarify how the ssp-operator works. About the patch per se, Yanir pretty much expressed my biggest/only concern here (https://github.com/MarSik/kubevirt-ssp-operator/pull/66#issuecomment-516463761)

Here's the script mentioned above: https://github.com/MarSik/kubevirt-ssp-operator/blob/master/build/make-manifests.sh

We use it when we do a release. The heavy lifting is performed by travis when we make a tag. 
1. we push a new tag
2. travis detects the tag and triggers a build
3. when we do a build reacting to a tag, we build the manifests among other things, using `make manifests`
4. a dependency of the `manifests` target is `operator-sdk`, which is downloaded from github if missing (always on travis builds)
5. when everything is set, we call `make manifests`, which in turn will leverage operator-sdk to do the work

Now, all this is needed to produce manifests to be consumed by HCO, so overall I think this patch is an improvement to our flow. Waiting to see the next upload.

**Comment by MarSik** [comment]
My issue is that HCO is leaking its implementation requirements to other
projects. I must say I do not like that at all.

One example of that is the golang code we have added just for HCO. From
this repository point of view it is dead code. And one we have to maintain
even though we already have CR in Yamil form.  The same applies to the
script you want to add. The operator does not need it for anything. Why
should it live here or be shipped with the operator then?

It is similar as if somebody decided to ship a Fedora specific template
spec file within an RPM that is built using it. It is inverting the proper
responsibility chain. Imagine what would happen if every consumer started
doing that. NMO for example can be useful even without HCO.

I simply can’t accept this way of presenting HCO requirements. Please stop
forcing every sub project to bend to HCO’s will and start talking to us
during the design phase. Imagine if every consumer started doing that.

I think the proper way would be to consume the CSV we already generate with
placeholders. We can update it if needed too. We can include the template
in the image even though I do not think it is the right approach. We can
even create a label that would reveal its location. But a HCO specific
templating script?

Why can’t you just reinvent distgit for HCO and collect the data from a
place everybody would post their images and manifests from. Is there a
reason for this inversion you are proposing?

And Yanir is right.  I started my vacation today so this couldn’t have
arrived at worse time. I will be back in about two weeks.

út 30. 7. 2019 v 16:00 odesílatel David Vossel <notifications@github.com>
napsal:

**Comment by davidvossel** [comment]
> Why can’t you just reinvent distgit for HCO and collect the data from a
place everybody would post their images and manifests from. 

We're using the build artifact of each operator release (the container image) to extract the CSV manifest. This is the standard the HCO requires for integration.

**Comment by djzager** [comment]
> My issue is that HCO is leaking its implementation requirements to other projects. I must say I do not like that at all.

Personally, I think there is a significant difference between implementation requirements and build + release requirements. I also don't think there is some broken inversion principle in wanting all of the components of a project to be on the same page when it comes to building and releasing the thing. OLM and CSVs are a gigantic pain, that isn't the HCO's fault, it's the hand we've been dealt and this is an attempt to mitigate that pain.

>  I think the proper way would be to consume the CSV we already generate with placeholders. We can update it if needed too. We can include the template in the image even though I do not think it is the right approach. We can even create a label that would reveal its location. But a HCO specific templating script?

I want to consider this option of putting, say a go template CSV + other OLM manifests, in the container image but "even though I do not think it is the right approach" gives me serious pause. Putting manifests in the container image is specifically the recommendation for OLM installed operators, putting templates on the container image is not a huge deviation from that. BUT, even if we did pursue this templated out CSV...we would still have to agree on a pattern for what the variables in that template would be so the dependency inversion problem you are talking about is still alive and well.

Doing a bunch of http requests on github release pages strikes me as a hack that I don't think is a real option worth considering.

> And Yanir is right. I started my vacation today so this couldn’t have arrived at worse time. I will be back in about two weeks.

I think the best thing that we could do to honor your being on vacation is to take a minute to consider what including templates in the container image would look like and how that would work.

File: build/make-csv-generator.sh [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+# cleanup placeholder version's deployment dir
+rm -rf mv deploy/olm-catalog/kubevirt-ssp-operator/${PLACEHOLDER_CSV_VERSION}
+
+# replace placeholder version with a human readable variable name
+# that will be used later on by csv-generator
+sed -i "s/${PLACEHOLDER_CSV_VERSION}/PLACEHOLDER_CSV_VERSION/g" $MANIFESTS_GENERATED_CSV 
+
+# inject the CRD and Description related data into the CSV
+cp $MANIFESTS_GENERATED_CSV ${MANIFESTS_GENERATED_CSV}.tmp
+./build/update-olm.py ${MANIFESTS_GENERATED_CSV}.tmp > ${MANIFESTS_GENERATED_CSV}
```
yanirq: @fromanirh  is there a chance update-olm.py with this PR change needs to be updated for every version, lets say if make some structural changes or naming ? (asking since you wrote it)

davidvossel: > I want to consider this option of putting, say a go template CSV + other OLM manifests, in the container image but "even though I do not think it is the right approach" gives me serious pause. Putting manifests in the container image is specifically the recommendation for OLM installed operators, putting templates on the container image is not a huge deviation from that. 

the image has a templated CSV yaml file that's being injected into it. the csv-generator script is  validating args and doing search/replace on that template, then spitting it to stdout.

MarSik:
> I think the best thing that we could do to honor your being on vacation is to take a minute to consider what including templates in the container image would look like and how that would work.

Actually, we can keep the script. Just do not make the path, name and arguments hardcoded, read them from a hco specific image label instead. That would make it easy to make this compatible with HCO while also allowing this to play nicely with the image structure itself. It won't fix the inversion, but it would make it bearable. Would that work for you?

davidvossel:
> Actually, we can keep the script. Just do not make the path, name and arguments hardcoded, read them from a hco specific image label instead. That would make it easy to make this compatible with HCO while also allowing this to play nicely with the image structure itself. It won't fix the inversion, but it would make it bearable. Would that work for you?

The only requirement we have is that there is a csv-generator script in /usr/bin/csv-generator that allows a CSV file to be generated via whatever arguments the csv-generator for the project in question requires. 

@yanirq @fromanirh 

any updates on the review? Did the new way I structured the PR fix your concerns with regards to duplication?

build/make-manifests.sh
```diff
@@ -53,13 +43,7 @@ for MF in deploy/service_account.yaml deploy/role.yaml deploy/role_binding.yaml
...
 ) > ${CLUSTER_VERSIONED_DIR}/kubevirt-ssp-operator.yaml
 
-# TODO: we should use --from-version
-# note: this creates under deploy/olm-catalog ...
-${OPERATOR_SDK} olm-catalog gen-csv --csv-version ${VERSION}
-
-./build/update-olm.py \
-	deploy/olm-catalog/kubevirt-ssp-operator/${VERSION}/kubevirt-ssp-operator.${TAG}.clusterserviceversion.yaml > \
-	${MANIFESTS_VERSIONED_DIR}/kubevirt-ssp-operator.${TAG}.clusterserviceversion.yaml
+./build/csv-generator.sh --csv-version=${VERSION} --namespace=placeholder --operator-image=REPLACE_IMAGE > ${MANIFESTS_VERSIONED_DIR}/kubevirt-ssp-operator.${TAG}.clusterserviceversion.yaml
```
yanirq: Just to see that were not missing something,  if we run make-manifests ,e.g. when releasing a tag - triggered by .travis.yaml : `make release "IMAGE_TAG=$TRAVIS_TAG"` 
where is the swap/input for `REPLACE_IMAGE` coming from ?

davidvossel: > where is the swap/input for REPLACE_IMAGE coming from ?

The release CSVs are still using the "REPLACE_IMAGE" image value. I don't know if that's intentional or not. I'm just replicating the existing behavior 1 for 1. 

For example, take a look at the v1.0.5 CSV (https://github.com/MarSik/kubevirt-ssp-operator/releases/download/v1.0.5/kubevirt-ssp-operator.v1.0.5.clusterserviceversion.yaml) It still has REPLACE_IMAGE in the released CSV. 

I noticed other odd things as well in the release process. For example the release kubevirt-ssp-operator.yaml for v1.0.5 (https://github.com/MarSik/kubevirt-ssp-operator/releases/download/v1.0.5/kubevirt-ssp-operator.yaml) is using the floating "latest" tag, which basically means there's no reason in making releases because they all point to the "latest" release container image regardless of the version of the manifest that is installed.

I didn't want to get involved in all that, so I just preserved whatever behavior currently exists for the ssp release process. For the purposes of the HCO, the csv-generator script lets us have the control needed to build the CSV accurately in the HCO's CSV build process regardless of what the ssp's release process is doing.

MarSik: Where do you see latest? REPLACE_IMAGE can easily be replaced with the hash or proper name:tag pair.

MarSik:
> The only requirement we have is that there is a csv-generator script in /usr/bin/csv-generator that allows a CSV file to be generated via whatever arguments the csv-generator for the project in question requires.

And I am politely asking for a small change on the HCO side to read the script name from a label, just so we can structure our image as we want and not as HCO dictates. Also the label could be versioned to allow for possible future "api" changes. Say if we added this to the Dockerfile:

```
LABEL org.kubevirt.hco.csv-generator.v1="/usr/bin/csv-generator"
```

Btw, while looking at that script.. how does HCO know what arguments are supposed to be passed here? Why not make this generic and pass all the container versions to all operators and let them decide if and how they want to use the data? I think that would simplify the HCO code (common code for all operators) and shift the responsibility to the proper people. Also it would make it easier to do changes (no commit to hco would be needed).

davidvossel:
> And I am politely asking for a small change on the HCO side to read the script name from a label, just so we can structure our image as we want and not as HCO dictates

I see. I don't have any issues with using a label to point to the location of csv-generator. 

>  Btw, while looking at that script.. how does HCO know what arguments are supposed to be passed here? Why not make this generic and pass all the container versions to all operators and let them decide if and how they want to use the data? I think that would simplify the HCO code (common code for all operators) and shift the responsibility to the proper people. Also it would make it easier to do changes (no commit to hco would be needed).

yeah, I'd really like that to be the case as well. I'm unsure about the feasibility of executing on that right now.

The issue is some operators want image tags, some want image SHAs, some want full images with repo image tag/sha, and some want args that aren't even images. It's a mixed bag right now. In order to get this initial effort in motion, the plan is for the HCO to absorb that complexity by taking all of the information it has and translate that info to whatever each component-level operator's csv-generator requires as arguments. I think that's just what has to be done to get this thing off the ground. 

Once the dust settles and we see what integration looks like in practice, we'll have a better understanding of what it would look like to converge on a common set of env vars that each csv-generator knows how to pull from.

build/make-manifests.sh
davidvossel: > Where do you see latest

in the kubevirt-ssp-operator.yaml manifest for each release. 

```
wget https://github.com/MarSik/kubevirt-ssp-operator/releases/download/v1.0.5/kubevirt-ssp-operator.yaml && cat kubevirt-ssp-operator.yaml | grep latest
.
.
.
          image: quay.io/fromani/kubevirt-ssp-operator-container:latest
```

MarSik: Ah, that is the manifest file with filled in data. This looks like a bug. @fromanirh

MarSik:
> I see. I don't have any issues with using a label to point to the location of csv-generator.

Great, then we have an agreement. I really believe this would make the design cleaner and more in accordance with who is responsible for the image.

> Once the dust settles and we see what integration looks like in practice, we'll have a better understanding of what it would look like to converge on a common set of env vars that each csv-generator knows how to pull from.

Yeah, the env vars method would make the templating really easy.

SSP wants the full image names including registry, name and tag or hash whatever is provided. Every operator should be consuming that tripple, because downstream versions might be using a different registry as well as different image names. Tag/hash is obviously needed as well even for upstream.

build/make-manifests.sh
fromanirh: Indeed REPLACE_IMAGE should be replaced (no pun intended) by our automation when we do release. @davidvossel not sure we should fix before or after your PR, whatever works best for you

fromanirh:
> > Actually, we can keep the script. Just do not make the path, name and arguments hardcoded, read them from a hco specific image label instead. That would make it easy to make this compatible with HCO while also allowing this to play nicely with the image structure itself. It won't fix the inversion, but it would make it bearable. Would that work for you?
> 
> The only requirement we have is that there is a csv-generator script in /usr/bin/csv-generator that allows a CSV file to be generated via whatever arguments the csv-generator for the project in question requires.
> 
> @yanirq @fromanirh
> 
> any updates on the review? Did the new way I structured the PR fix your concerns with regards to duplication?

Sorry, I had to take care of other tasks - including another 1.0.5 release, like you noticed, which we used as checkpoint release before merging this PR.

Good news is Martin/Yanir already expressed 80% of my points, so we are in a good track. I'll review and play with this PR today (20190801), so expect progress here.

build/make-csv-generator.sh
fromanirh: @yanirq we indeed need to do some changes to the update-olm.py script - but this seems to be needed only when add new CRDs, like when I just did in v1.0.5 and missed that. Nothing else seems to be needed.
If nothing else, this PR is a good chance to review our automation and (admittedly messy) flow and simplify them

build/csv-generator.sh
```diff
@@ -0,0 +1,112 @@
...
+    exit 1
+fi
+
+cp ${MANIFESTS_GENERATED_CSV} ${TMP_FILE}
+
+# replace placeholder version with a human readable variable name
+# that will be used later on by csv-generator
+sed -i "s/PLACEHOLDER_CSV_VERSION/${CSV_VERSION}/g" ${TMP_FILE}
+sed -i "s/namespace: placeholder/${NAMESPACE}/g" ${TMP_FILE}
+sed -i "s/REPLACE_IMAGE/${OPERATOR_IMAGE}/g" ${TMP_FILE}
```
fromanirh: does this deal properly with slashes in the image path? example:
```
$ ./build/csv-generator.sh --csv-version='1.2.0' --namespace='openshift-cnv' --operator-image='quay.io/fromani/kvc' > /dev/null 
sed: -e expression #1, char 25: unknown option to `s'
```

compare with
```
$ ./build/csv-generator.sh --csv-version='1.2.0' --namespace='openshift-cnv' --operator-image='foobar' > /dev/null 
$ echo $?
0
```

fromanirh:
Here's the fix for the issues not related to this PR but discovered here. https://github.com/MarSik/kubevirt-ssp-operator/pull/69 . I'm fine rebasing that PR on top of this one.

build/make-manifests.sh
fromanirh: For reference:
operator image should be
1. on travis, since we are going to test the image out of the sources BEFORE the upload on quay.io:
`quay.io/fromani/kubevirt-ssp-operator-container:$TAG`, where `tag` could be in turn `devel` or `$TRAVIS_TAG`, aka the current tag
2. everywhere else, the only thing which changes is the tag, which should be the last released tag, found using github APIs or any other means.

build/csv-generator.sh
davidvossel: oh great catch!

davidvossel:
@fromanirh I believe all the feedback as been addressed.

I added the
* Docker LABEL pointing to location of csv-generator
* Ability to dump CRDs with csv-generator (something I missed that is required for HCO integration)
* Fixes issue with image names that have slashes in them

davidvossel:
thanks everyone!

PR #67, opened 2019-07-30
roles: add role to deploy metric aggregation rules

fromanirh:
Using PrometheusRules to configure the aggregation rules,
we can create the metrics CNV wants to consume out of the
metrics that kubevirt provides out of the box.

### DIFF
--- a/deploy/crds/kubevirt_v1_metricsaggregation_cr.yaml
+++ b/deploy/crds/kubevirt_v1_metricsaggregation_cr.yaml
@@ -1,0 +1,6 @@
+apiVersion: kubevirt.io/v1
+kind: KubevirtMetricAggregation
+metadata:
+  name: kubevirt-metrics-aggregation
+spec:
+  version: v0.0.1
--- a/deploy/crds/kubevirt_v1_metricsaggregation_crd.yaml
+++ b/deploy/crds/kubevirt_v1_metricsaggregation_crd.yaml
@@ -1,7 +1,7 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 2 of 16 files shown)


roles/KubevirtMetricsAggregator/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# defaults file for KubevirtCommonTemplatesBundle
```
fromanirh: rename

roles/KubevirtMetricsAggregator/templates/aggregation-rule-vmi-count.yaml.j2
```diff
@@ -0,0 +1,15 @@
...
+    prometheus: k8s
+    role: cnv-rules
+  name: prometheus-k8s-rules-cnv
+  namespace: "{{ meta.namespace }}"
+spec:
+  groups:
+  - name: cnv.rules
+    rules:
+    - expr: |
+        sum(kubevirt_vmi_phase_count{phase="running}) by (node)
```
fromanirh: typo

ksimon1:
LGTM
just one thing, do we need these files? 
roles/KubevirtMetricsAggregation/defaults/main.yml
roles/KubevirtMetricsAggregation/meta/main.yml
roles/KubevirtMetricsAggregation/handlers/main.yml

fromanirh:
> LGTM
> just one thing, do we need these files?
> roles/KubevirtMetricsAggregation/defaults/main.yml
> roles/KubevirtMetricsAggregation/meta/main.yml
> roles/KubevirtMetricsAggregation/handlers/main.yml

We don't, but not sure the ansible role layout needs them. If not we can cleanup this and other roles, will do in a separate patch.

PR #69, opened 2019-08-01
manifests: fix operator image

fromanirh:
We need to fix various flows on which we build manifests:
- the release flow (entry point: make-manifests.sh)
- the install flow (entry point: install-operator.sh)

In any case: we should never leave the placeholder `REPLACE_IMAGE` in
the manifests fail, and we should never use floating tags ('devel',
'latest'...)  outside the travis *testing* stage.


fromanirh:
/cc @ksimon1 @MarSik 
not really urgent

PR #70, opened 2019-08-01
Fix olm aggregation rules

fromanirh:
PR https://github.com/MarSik/kubevirt-ssp-operator/pull/67 had a couple bugs with previous testing (on OCP 3.11) didn't reveal, and lacked proper OLM support. This PR aims to fill the gaps.


fromanirh:
waiting for a full round of testing on OKD >= 4.1

fromanirh:
verified that with this patch the cr makes the operator correctly deploy the rule on okd 4.1.0
Can't test yet the flow end 2 end, but I think it's good enough for this PR

PR #75, opened 2019-08-12
conditions func-tests for node-labeller and template-validator

ksimon1:
This PR adds condition func test for template-validator and node-labeller

### DIFF
--- a/functests/08-test-conditions-node-labeller.sh
+++ b/functests/08-test-conditions-node-labeller.sh
@@ -8,26 +8,25 @@ source ${SCRIPTPATH}/testlib.sh
 RET=0
 TEST_NS="${KV_NAMESPACE}"
 
-exit 0
-
 oc create -n ${TEST_NS} -f "${SCRIPTPATH}/node-labeller-unversioned-cr.yaml" || exit 2
 # TODO: SSP-operator needs to improve its feedback mechanism
 # fetching node-labeller images may take a while
 
 wait_node_labeller_running ${TEST_NS} 5 60
+
 #wait for ssp operator to set proper conditions
 sleep 15s
 
-if [ $(oc get KubevirtNodeLabellerBundle -o json | jq  '.items[0].status.conditions[] | select((.type=="Progressing") and (.status="True")) | .type' | wc -l) -eq 0 ]; then
-		RET=0
-fi
-
 if [ $(oc get KubevirtNodeLabellerBundle -o json | jq  '.items[0].status.conditions[] | select((.type=="Progressing") and (.status="False")) | .type' | wc -l) -eq 0 ]; then
-		RET=0
+	echo "Progressing condition is not set to false"
+	RET=1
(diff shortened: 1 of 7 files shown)


functests/08-test-conditions-node-labeller.sh
```diff
@@ -0,0 +1,35 @@
...
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/node-labeller-unversioned-cr.yaml" || exit 2
+# TODO: SSP-operator needs to improve its feedback mechanism
+# fetching node-labeller images may take a while
+
+wait_node_labeller_running ${TEST_NS} 5 60
+#wait for ssp operator to set proper conditions
+sleep 15s
+
+if [ $(oc get KubevirtNodeLabellerBundle -o json | jq  '.items[0].status.conditions[] | select((.type=="Progressing") and (.status="True")) | .type' | wc -l) -eq 0 ]; then
+		RET=0
```
fromanirh: the logic is confusing to me. You should either:
- assume the test will run OK, check conditions, bail out with failure at first unexpected state and return error
- assume the test will fail, check conditions, only iff all of them are the expected state then return success
You seem to do a strange mix of the two in the code

functests/09-test-conditions-template-validator.sh
```diff
@@ -0,0 +1,34 @@
...
+
+SCRIPTPATH=$( dirname $(readlink -f $0) )
+source ${SCRIPTPATH}/testlib.sh
+
+[ -z ${KV_NAMESPACE} ] && exit 127
+
+RET=0
+TEST_NS="${KV_NAMESPACE}"
+
+exit 0
```
fromanirh: probably unneeded

ksimon1:
@MarSik can you please review this?

roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
 # defaults file for KubevirtNodeLabeller
 wait: true
+use_kvm: true
```
fromanirh: do we want to have this change in the node-labeller upstream/standalone project?
if not, we have a fork here for all intents and purposes (which could be perfectly fine as long we acknowledge this).

ksimon1: node-labeller already contains this logic.

fromanirh:
> Looks good, and, looking at the tip of the branch, better than before. I'm pretty sure that travis tests didn't fail _all the times_, though.

Answering to myself: that was because _someone_ cut some corners: 7fc7644f519127d2ab4afb496f94253fe7aba122

Nice fix, let's go with it.

roles/KubevirtNodeLabeller/defaults/main.yml
fromanirh: Great then.

PR #76, opened 2019-08-14
CR conditions refactored

ksimon1:
This PR contains refactored ssp operator conditions.

This PR Removes degraded condition for common templates - in this state it doesn't make sense to have it there,
because in every iteration all common templates will be installed.

This PR contains k8s_status.py file, which fixes bugs in k8s_status ansible module:
1) When CR condition is on index 0, then new condition with same type should replace old one.
But instead this new condition is appended to list of conditions.
2) object_contains function compares if new conditions are subset of old conditions, but due to lastTransitionTime
they are never subset. New conditions should not contain lastTransitionTime too. It will be set by kubernetes.
/cc @MarSik @fromanirh

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -10,7 +10,9 @@ COPY _defaults.yml ${HOME}/_defaults.yml
 COPY deploy/crds ${HOME}/deploy/crds
 COPY manifests/generated ${HOME}/manifests/generated
 COPY build/csv-generator.sh /usr/bin/csv-generator
-#original module k8s_status contains bug, which prevents ssp operator to correctly set conditions
+#original module k8s_status contains bugs, which prevents ssp operator to correctly set conditions
+#https://github.com/operator-framework/operator-sdk/pull/1820
+#https://github.com/operator-framework/operator-sdk/pull/1845
 COPY build/k8s_status.py /usr/share/ansible/openshift
 
 COPY build/preprocess_template.py ${HOME}/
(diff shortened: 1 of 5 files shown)


build/Dockerfile
```diff
@@ -10,6 +10,8 @@ COPY _defaults.yml ${HOME}/_defaults.yml
 COPY deploy/crds ${HOME}/deploy/crds
 COPY manifests/generated ${HOME}/manifests/generated
 COPY build/csv-generator.sh /usr/bin/csv-generator
+#original module k8s_status contains bug, which prevents ssp operator to correctly set conditions
+COPY build/k8s_status.py /usr/share/ansible/openshift
```
fromanirh: please mention the u/s k8s_status.py bug/issue so we can track it and drop this workaround as soon as possible

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -16,135 +16,26 @@
   set_fact:
     deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template') }}"
 
-- name: "Get common-templates-cr"
```
fromanirh: I want this change, and it looks good at glance. But it's hard to see how the old, convoluted code (because of k8s_status bugs) translatest to the new lean and mean one. Thus, please
- add comments/or amend the commit message to explain in a couple lines how the fixes in the k8s_status allow to simplify the logic (e.g. why it was convoluted before?)
- explicitely highlight if the new conditions refactored are 1:1 simplifications of the old logic or if you dropped some conditions, as it is hinted in the current commit message. E.g: which conditions where reported before? what now?

fromanirh:
waiting for CI, then we can marge

PR #77, opened 2019-08-19
Fix dockerfile.registry to work with origin-operator-registry 4.2.0

ksimon1:
Latest build of registry container fails with error: 
time="2019-08-19T10:13:44Z" level=fatal msg="permissive mode disabled" error="error loading manifests from directory: could not decode contents of file /registry/bundles.db into package: error converting YAML to JSON: yaml: control characters are not allowed"

This PR fixes this error (same issue as here: https://github.com/kubevirt/cluster-network-addons-operator/pull/215).

/cc @MarSik, @fromanirh

### DIFF
--- a/build/Dockerfile.registry
+++ b/build/Dockerfile.registry
@@ -1,10 +1,13 @@
-FROM quay.io/openshift/origin-operator-registry
+FROM quay.io/openshift/origin-operator-registry:4.2.0
 
-COPY manifests /registry
+COPY manifests manifests
+
+# Bundle should include only manifests related to the CSV,
+# drop all additional files.
+RUN find manifests/kubevirt-ssp-operator/*/ -type f ! -name '*_crd.*' ! -name '*.clusterserviceversion.*' -exec rm -f {} \;
 
 # Initialize the database
-RUN rm -rf /registry/generated && \
-    initializer --manifests /registry --output bundles.db
+RUN initializer --manifests manifests/kubevirt-ssp-operator/ --output bundles.db
 
 # There are multiple binaries in the origin-operator-registry
 # We want the registry-server
--- a/build/Dockerfile.registry
+++ b/build/Dockerfile.registry
@@ -4,7 +4,7 @@ COPY manifests manifests
 
 # Bundle should include only manifests related to the CSV,
 # drop all additional files.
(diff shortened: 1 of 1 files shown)


MarSik:
The fix itself looks good, but I am a bit uncertain about the maintainability. Do we expect to need different manifest bundle for different openshift versions? In other words, should this be configurable somehow?

build/Dockerfile.registry
```diff
@@ -1,10 +1,13 @@
-FROM quay.io/openshift/origin-operator-registry
+FROM quay.io/openshift/origin-operator-registry:4.2.0
 
-COPY manifests /registry
+COPY manifests manifests
```
fromanirh: I'm not sure thus works OK with HCO/registry bundles. The filesystem layout is significant here.

build/Dockerfile.registry
```diff
@@ -1,10 +1,13 @@
...
+
+# Bundle should include only manifests related to the CSV,
+# drop all additional files.
+RUN find manifests/kubevirt-ssp-operator/*/ -type f ! -name '*_crd.*' ! -name '*.clusterserviceversion.*' -exec rm -f {} \;
```
fromanirh: this works but I'm not too happy with it. We should log spurious/unexpected entries.

build/Dockerfile.registry
```diff
@@ -1,10 +1,13 @@
...
 
 # Initialize the database
-RUN rm -rf /registry/generated && \
-    initializer --manifests /registry --output bundles.db
+RUN initializer --manifests manifests/kubevirt-ssp-operator/ --output bundles.db
```
fromanirh: same concern as per line #3

build/Dockerfile.registry
fromanirh: My source is just https://github.com/operator-framework/operator-registry

build/Dockerfile.registry
fromanirh: maybe just use the `-v flag` of the `rm` command?

build/Dockerfile.registry
ksimon1: according to this https://github.com/operator-framework/operator-registry/blob/master/cmd/registry-server/main.go#L36 registry server cares only about bundles.db path

PR #84, opened 2019-08-29
Conditions fixes

fromanirh:
collection of fixes about condition reporting


fromanirh:
works as expected, verified on top of git master manually checking the cluster (pods, conditions, logs)

PR #89, opened 2019-09-05
manifests: declare storage requests

fromanirh:
Looking at the development clusters,  we see he pod requires >= 700M of storage.
We intentionally overestimate to 1G to have some safety.

Related-To: https://bugzilla.redhat.com/1748873

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -25,6 +25,9 @@ spec:
           - containerPort: 60000
             name: metrics
           imagePullPolicy: Always
+          resources:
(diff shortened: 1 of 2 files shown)


fromanirh:
we can't test this on our CI, which runs on minishift:
```
message: '0/1 nodes are available: 1 Insufficient ephemeral-storage.
```

fromanirh:
Looks like we don't really need this change to solve the eviction issue, and on top of it looks like `ephemeral-storage` is not supported in our CI.

PR #91, opened 2019-09-16
adds test for available condition for common templates

ksimon1:
This PR adds test which is testing available condition for common templates

### DIFF
--- a/functests/10-test-conditions-templates.sh
+++ b/functests/10-test-conditions-templates.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+SCRIPTPATH=$( dirname $(readlink -f $0) )
+source ${SCRIPTPATH}/testlib.sh
+
+[ -z ${KV_NAMESPACE} ] && exit 127
+
+RET=1
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${KV_NAMESPACE} -f "${SCRIPTPATH}/common-templates-versioned-cr.yaml" || exit 2
+
+wait_for_deployment_ready ${TEST_NS} 5 50 "KubevirtCommonTemplatesBundle" "Running" "Successful"
+RET="$?"
+if [ $RET -eq 1 ]; then
+    exit $RET
+fi
+
+for idx in $( seq 1 40); do
+	NUM=$(oc get templates -n ${TEST_NS} -l "template.kubevirt.io/type=base" -o json | jq ".items | length")
+	echo "templates found in ${TEST_NS}: ${NUM}"
+	if [ "${NUM}" > 0 ]; then
+		break
+	fi
+	sleep 3s
+done
+
+#wait for ssp operator to set proper conditions
+wait_for_condition ${TEST_NS} 5 40 "KubevirtCommonTemplatesBundle" "Available" "True"
+RET="$?"
+
+oc delete -n ${TEST_NS} -f "${SCRIPTPATH}/common-templates-versioned-cr.yaml" || exit 2
+
+exit $RET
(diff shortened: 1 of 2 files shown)


functests/10-test-conditions-templates.sh
```diff
@@ -0,0 +1,26 @@
...
+source ${SCRIPTPATH}/testlib.sh
+
+[ -z ${KV_NAMESPACE} ] && exit 127
+
+RET=1
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${KV_NAMESPACE} -f "${SCRIPTPATH}/common-templates-versioned-cr.yaml" || exit 2
+
+# 76 is number of all templates in version v0.6.0
```
MarSik: I am not sure we should hardcode this number. I will require maintenance and is pretty well hidden. Is there any other way?

ksimon1: I don't like it either. I can try take version from CR and then e.g. count how many templates are in the file.

MarSik: What about reading the ansible run status from the CR itself? If the ansible finished, then all templates must surely be there.

ksimon1: good idea

functests/testlib.sh
```diff
@@ -99,10 +99,31 @@ wait_for_condition() {
 	local status=$6
 	for num in $( seq 1 ${max_tries} ); do
 		(( ${V} >= 1 )) && echo "waiting for $type condition availability: ${num}/${max_tries}"
-		if [ $(oc get ${kind} -o json ${NS} | jq '.items[0].status.conditions[] | select((.type=="'${type}'") and (.status="'${status}'"))' | wc -l) -gt 0 ]; then
+		if [ $(oc get ${kind} -o json ${NS} | jq 'select(.items[0].status.conditions != null) | .items[0].status.conditions[] | select((.type=="'${type}'") and (.status="'${status}'"))' | wc -l) -gt 0 ]; then
```
fromanirh: could you please explain why do we need this change?

functests/10-test-conditions-templates.sh
```diff
@@ -0,0 +1,21 @@
...
+
+RET=1
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${KV_NAMESPACE} -f "${SCRIPTPATH}/common-templates-versioned-cr.yaml" || exit 2
+
+wait_for_deployment_ready ${TEST_NS} 5 50 "KubevirtCommonTemplatesBundle" "Running" "Successful"
+
+#wait for ssp operator to set proper conditions
+wait_for_condition ${TEST_NS} 5 20 "KubevirtCommonTemplatesBundle" "Available" "True"
```
fromanirh: do we want to add a test that *some* templates are actually there? Something like template count > 1. That should cover all corners.

MarSik: We could do that too, but we need to use the right label (template.kubevirt.io: base iirc) to avoid seeing all the openshift ones.

functests/testlib.sh
ksimon1: When CR is newly created, it doesn't contain conditions array and due to that, jq complains:
`jq: error (at <stdin>:26): Cannot iterate over null (null)`

MarSik: What about using:

> Alternative operator: //
>
> A filter of the form a // b produces the same results as a, if a produces results other than false and null. Otherwise, a // b produces the same results as b.
> 
> This is useful for providing defaults: .foo // 1 will evaluate to 1 if there’s no .foo element in the input. It’s similar to how or is sometimes used in Python (jq’s or operator is reserved for strictly Boolean operations).

PR #92, opened 2019-09-16
defaults: consolidate in ./_defaults.yml

fromanirh:
We should have a single place on which we set the defaults,
and that place should be the toplevel _defaults.yml.
We should use per-role default only in rare, exceptional, well
documented cases, which we don't have now.

### DIFF
--- a/_defaults.yml
+++ b/_defaults.yml
@@ -2,21 +2,12 @@
 # role values. This is useful for "downstream"
 # packaging and opinionated releases.
 #
-# If in doubt, leave the file empty.
-#
-# Example:
-#
-# validator_namespace: myproject
-# validator_image: registry.example.com/mine/kubevirt-template-validator
-# validator_tag: v0.6.1
-# templates_version: v0.6.2
+# this file should contain only the tags (~= versions) of the containers to install.
+# every other package/role specific value should sit in `roles/*/defaults/main.yml`.
 
----
-#node-labeller
 kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.1', true) }}"
 kvm_info_nfd_plugin_tag: "{{ lookup('env','KVM_INFO_TAG')| default('v0.5.8', true) }}"
 kubevirt_cpu_nfd_plugin_tag: "{{ lookup('env','CPU_PLUGIN_TAG')| default('v0.1.1', true) }}"
-virt_launcher_tag: "{{ lookup('env','VIRT_LAUNCHER_TAG')| default('v0.19.0', true) }}"
-
-#validator
+virt_launcher_tag: "{{ lookup('env','VIRT_LAUNCHER_TAG')| default('v0.21.0', true) }}"
 validator_tag: "{{ lookup('env','VALIDATOR_TAG')| default('v0.6.2', true) }}"
+templates_version: v0.7.0
--- a/_defaults.yml
+++ b/_defaults.yml
@@ -9,6 +9,13 @@
 #
 # A notable exception are common templates, on which we use a two-steps approach to allow the
 # users to easily override the installed version. Templates need to be special.
+#
+# When possible, the 'version' value in the CRs should take precedence over the defaults value.
+# The authoritative sources are, in decreasing priority order
+# 1. spec.version in the CRs (set by users)
+# 2. the *_TAG environment veriable (set by HCO)
+# 3. the builtin defaults (set by devs/operator)
+# Any version, or default, should not be set more than once in a place in the source tree.
 
 kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.1', true) }}"
 kvm_info_nfd_plugin_tag: "{{ lookup('env','KVM_INFO_TAG')| default('v0.5.8', true) }}"
--- a/_defaults.yml
+++ b/_defaults.yml
@@ -2,7 +2,9 @@
 # role values. This is useful for "downstream"
 # packaging and opinionated releases.
 #
-# this file should contain only the tags (~= versions) of the containers to install.
+# This file can be left empty to use the latest and greatest versions of all containers.
+# Or it can contain the tags (~= versions) of the containers to install if there is a known combination you want.
+# In case you really know what you are doing it is also possible to override the container names here.
 # every other package/role specific value should sit in `roles/*/defaults/main.yml`.
 
 kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.1', true) }}"
--- a/_defaults.yml
+++ b/_defaults.yml
@@ -15,6 +15,9 @@
 # 1. spec.version in the CRs (set by users)
 # 2. the *_TAG environment veriable (set by HCO)
 # 3. the builtin defaults (set by devs/operator)
+# 4. the per-role defaults, which is the last-resort value. Use a low-maintenance, usually working,
+#    easy to spot value. 'latest' is a good value here - probably the only case where we should use it.
+#
 # Any version, or default, should not be set more than once in a place in the source tree.
 
 kubevirt_node_labeller_tag: "{{ lookup('env','NODE_LABELLER_TAG')| default('v0.1.1', true) }}"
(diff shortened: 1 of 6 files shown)


roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -2,12 +2,3 @@
 # defaults file for KubevirtNodeLabeller
 wait: true
 use_kvm: true
-
-kubevirt_node_labeller_image: "node-labeller"
```
MarSik: I like it when the role is working when executed manually. I commonly treat the defaults file as docs for the role too. So personally I would keep generic defaults here.

roles/KubevirtTemplateValidator/defaults/main.yml
```diff
@@ -1,6 +1,3 @@
 ---
 # defaults file for KubevirtTemplateValidator
 wait: true
-validator_image: kubevirt-template-validator
-# this comes from the CR, we must diverge from validator_$SOMETHING standard naming
-version: "{{ validator_tag |default('v0.6.2') }}"
```
MarSik: Here I would just use version: latest, but I would keep it here.

roles/KubevirtNodeLabeller/defaults/main.yml
fromanirh: OK, so move back the image names per-role and have the versions in the top-level `_defaults.yml`, plus adding few lines of doc explaining how it will work (= summarizing the discussion we're having here). Would that work?

roles/KubevirtTemplateValidator/defaults/main.yml
fromanirh: ack, see above

roles/KubevirtNodeLabeller/defaults/main.yml
MarSik: Yeah, keep the image names and versions here, but change all versions here to latest and explain that the vars here are just for manual execution purposes. Release related defaults should be set in the _defaults.yml file.

fromanirh: ack, done

roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -2,12 +2,7 @@
...
-kubevirt_node_labeller_tag: "latest"
-kvm_info_nfd_plugin_image: "kvm-info-nfd-plugin"
-kvm_info_nfd_plugin_tag: "latest"
-kubevirt_cpu_nfd_plugin_image: "cpu-nfd-plugin"
-kubevirt_cpu_nfd_plugin_tag: "latest"
-virt_launcher_image: "virt-launcher"
-virt_launcher_tag: "latest"
+kubevirt_node_labeller_image: kubevirt-cpu-node-labeller
```
ksimon1: Why you add kubevirt- prefix to all image names? It will break whole node-labeller role, because these images don't exists on kubevirt quay.

_defaults.yml
```diff
@@ -2,21 +2,12 @@
...
 #
-# If in doubt, leave the file empty.
-#
-# Example:
-#
-# validator_namespace: myproject
-# validator_image: registry.example.com/mine/kubevirt-template-validator
-# validator_tag: v0.6.1
-# templates_version: v0.6.2
+# this file should contain only the tags (~= versions) of the containers to install.
```
MarSik: What do would you say to this?

```suggestion
# This file can be left empty to use the latest and greatest versions of all containers.
# Or it can contain the tags (~= versions) of the containers to install if there is a known combination you want.
# In case you really know what you are doing it is also possible to override the container names here.
```

roles/KubevirtCommonTemplatesBundle/defaults/main.yml
```diff
@@ -1,4 +1,2 @@
 ---
 # defaults file for KubevirtCommonTemplatesBundle
-version: "{{ templates_version |default('v0.6.2') }}"
```
MarSik: This has to stay here. The CR and the playbook use the `version` field and the default here is used when the user omits it.

roles/KubevirtTemplateValidator/defaults/main.yml
```diff
@@ -2,5 +2,3 @@
 # defaults file for KubevirtTemplateValidator
 wait: true
 validator_image: kubevirt-template-validator
-# this comes from the CR, we must diverge from validator_$SOMETHING standard naming
-version: "{{ validator_tag |default('v0.6.2') }}"
```
MarSik: Same here. The CR exposes the version using the `version` variable. And the defaults have to follow the suit. We can replace the default with latest, but that is it.

fromanirh: In this case I think we should just omit the version from the CR we ship. This duplication seems harmful. We should instead document/validate the schema of the CR (in another PR ofc). Maybe also for the templates?

MarSik: We could omit the version from the validator CR. There is a good reason to have it for templates though, it allows you to install older versions of templates in case you need the older definition.

fromanirh: ack, works for me.

roles/KubevirtCommonTemplatesBundle/defaults/main.yml
fromanirh: ack, templates are special enough.

roles/KubevirtTemplateValidator/tasks/main.yml
```diff
@@ -1,5 +1,8 @@
 ---
 # tasks file for KubevirtTemplateValidator
+- name: Figure out validator version
+  set_fact:
+    validator_version: "{{ version | default(validator_tag, true) }}"
```
MarSik: You can do this directly in defaults/main.yml or in vars/main.yml, that way it stays local to the role. set-fact sets a global.

fromanirh: ok, let's do this in `defaults/main.yml`

roles/KubevirtCommonTemplatesBundle/defaults/main.yml
```diff
@@ -1,4 +1,3 @@
 ---
 # defaults file for KubevirtCommonTemplatesBundle
-version: "{{ templates_version |default('v0.7.0') }}"
-
+version: "{{ templates_version |default('v0.6.2') }}"
```
MarSik: I think you can remove the default() part here.

roles/KubevirtTemplateValidator/defaults/main.yml
```diff
@@ -2,5 +2,4 @@
...
+validator_version: "{{ version | default(validator_tag, true) }}"
```
MarSik: Why not use the same like in templates?

version: "{{ validator_tag }}"

It is already a defaults file and should behave like we want. Then just use {{ version }} in the tasks.

roles/KubevirtCommonTemplatesBundle/defaults/main.yml
fromanirh: done

roles/KubevirtTemplateValidator/defaults/main.yml
fromanirh: It's just me being inexperienced in ansible. Done.

roles/KubevirtTemplateValidator/defaults/main.yml
```diff
@@ -2,5 +2,4 @@
 # defaults file for KubevirtTemplateValidator
 wait: true
 validator_image: kubevirt-template-validator
-# this comes from the CR, we must diverge from validator_$SOMETHING standard naming
-version: "{{ validator_tag |default('v0.6.2') }}"
+validator_version: "{{ validator_tag }}"
```
MarSik: I kind of iterated back to beginning... I think this should be like follows:

```suggestion
validator_version: "{{ validator_tag |default('latest') }}"
```

Or we could name the variable as `version` in case we want to allow CR override. The role should then use just the name like you changed it.

fromanirh: works for me.

PR #94, opened 2019-09-17
Fix ssp operator image permissions

ksimon1:
This PR adds new permission to deploy manifest of ssp operator.
This new permission is needed to upgrade version of operator-sdk image.
//cc @MarSik, @fromanirh 

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -48,10 +48,12 @@ rules:
   - apps
   resources:
   - deployments
+  - deployments/finalizers
   - replicasets
   - daemonsets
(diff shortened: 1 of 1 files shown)


MarSik:
Just a note, please write a bit more specific commit message about what is actually the reason for fix. 

For example:

"""
Add update and deployment/finalizers permissions

This is needed for successful deployment of the Metric role when using operator-sdk X.Y. (Ideally link to any explanation for the change in operator sdk - issue number, commit, ... but not necessary if you can't find it).
"""

PR #95, opened 2019-09-17
implement a circuitbreaker-like utility role

fromanirh:
This PR implements a circuitbreaker-like pattern for the SSP operator.
It is meant to be used very rarely, and only in the-sky-is-falling scenarios, when everything else failed.
For this reason the documentation is intentionally kept minimal, and the advertisement of this feature should be almost non-existant.

This PR want sto use the new role only in the component that's more likely to trigger a cascading failure: the template validator.

### DIFF
--- a/roles/KubevirtCircuitBreaker/tasks/main.yml
+++ b/roles/KubevirtCircuitBreaker/tasks/main.yml
@@ -1,0 +1,14 @@
+---
+# tasks file for KubevirtCircuitBreaker
+# 
+#
+- name: Extract the operator POD info
+  set_fact:
+    operator_pod: "{{ lookup('k8s', api_version='v1', kind='Pod', label_selector='name=kubevirt-ssp-operator') | from_yaml }}"
+- name: Extract the disable info
+  set_fact:
+    current_breaker: "{{ operator_pod['metadata'].get('annotations', {})['kubevirt.io/emergency-disable'] | default('[]', true) | from_json }}"
+    current_cr: "{{ meta.namespace }}/{{ meta.name }}"
+- name: Emergency disable check
+  meta: end_play
+  when: current_cr in current_breaker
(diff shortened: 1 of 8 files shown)


fromanirh:
RFC @MarSik @ksimon1 @yanirq

fromanirh:
OK, after a few missteps the ansible-operator now properly digest the role. But it seems the annotation is just ignored :)

fromanirh:
Seems to work, but needs functests.

roles/KubevirtCircuitBreaker/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+# tasks file for KubevirtCircuitBreaker
+# 
+#
+- name: Extract the operator POD info
+  set_fact:
+    operator_pod: "{{ lookup('k8s', api_version='v1', kind='Pod', label_selector='name=kubevirt-ssp-operator') | from_yaml }}"
```
MarSik: I am a bit worried about the hardcoded label. Is there any way to introspect this?

fromanirh: We need to check. But we need also to update the `KubevirtRepoInfo` role.

fromanirh: Looks like there is no easy way to go up the chain CR -> deployment -> POD besides juggling with labels, which doesn't seem much of an improvement over the current CR.
On top of that, this is a bit hard to do in a generic way.
So if we fill it's a real concern we should tackle in a separate effort/PR

MarSik: Ack. The other option is to move the circuit breaker back to the CR. I wonder which direction is better.

fromanirh:
another approach: https://github.com/MarSik/kubevirt-ssp-operator/pull/97

fromanirh:
we decided to go forward in https://github.com/MarSik/kubevirt-ssp-operator/pull/97

PR #97, opened 2019-09-25
operator: implement circuitbreaker

fromanirh:
Implement circuitbreaker-like behaviour for the SSP operator.
With is change is now possible to *temporarily* make the operator ignore
one or more operands.
This is intended to be used as extrema ratio parachute
to troubleshoot issue in clusters, when everything else failed and the
state has to be recovered, or at very least investigated, somehow.

see `TROUBLESHOOTING.md` to learn about how to use this
functionality.


fromanirh:
tested on a live cluster. Works as expected. Not sure about meanongful functests, but they can be added later.

roles/KubevirtCircuitBreaker/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# defaults file for KubevirtCircuitBreaker
```
MarSik: Delete the empty files please.

fromanirh: Done (here and everywhere else)

fromanirh:
tested on a live cluster even after the last changes. Works as intended.

PR #100, opened 2019-09-26
playbooks: enable circuitbreaker for all our CRs

fromanirh:
We have use cases (on dev/QE envs) to pause the
reconciliations of all our CRs, so we enable
the role for all of them.

### DIFF
--- a/playbooks/commontemplatesbundle.yaml
+++ b/playbooks/commontemplatesbundle.yaml
@@ -2,6 +2,9 @@
   vars_files:
     - _defaults.yml
   roles:
+  - role: KubevirtCircuitBreaker
+    vars:
+    - cr_kind: KubevirtCommonTemplatesBundle
 # we want to prepare everything to pull containers from the same repo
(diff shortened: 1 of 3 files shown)


playbooks/commontemplatesbundle.yaml
```diff
@@ -2,6 +2,9 @@
   vars_files:
     - _defaults.yml
   roles:
+  - role: KubevirtCircuitBreaker
+    vars:
+    - cr_kind: KubevirtCommonTemplatesBundle
```
fromanirh: actually not sure about this one

playbooks/metricsaggregation.yaml
```diff
@@ -3,4 +3,7 @@
   vars_files:
     - _defaults.yml
   roles:
+  - role: KubevirtCircuitBreaker
+    vars:
+    - cr_kind: KubevirtMetricsAggregation
```
fromanirh: I got a use case for this while working on https://github.com/MarSik/kubevirt-ssp-operator/pull/98

PR #101, opened 2019-09-30
metrics aggregation: add kubevirt.io label

fromanirh:
And use it to unbreak the functest.


fromanirh:
requires manual testing - no prometheus deployed on CI atm, too resource hungry.

PR #105, opened 2019-10-08
Increase number of validator replicas to 2.

ksimon1:
User can set number of replicas via template-validator CR.

Kubevirt is using the same number of replicas for e.g. virt-controller (https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-operator/creation/components/deployments.go#L157).

@MarSik, @fromanirh what about HCO? Should it have ability to set replicas as well?


deploy/crds/kubevirt_v1_templatevalidator_cr.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
   namespace: kubevirt
 spec:
   version: v0.6.2
+  template_validator_replicas: 2
```
fromanirh: we need to settle for either snake_case or camelCase. Let's just stick to whatever kubernetes prefers, AFAIK camelCase

roles/KubevirtTemplateValidator/defaults/main.yml
```diff
@@ -3,3 +3,5 @@
 wait: true
 validator_image: kubevirt-template-validator
 validator_version: "{{ validator_tag | default('latest') }}"
+replicas: "{{ template_validator_replicas | default(2) }}"
```
fromanirh: this is ok because it is a "private" detail of the validator deployment, hence should sit here and not in the top-level `_defaults.yml`

PR #106, opened 2019-10-09
rename use_kvm to useKVM

ksimon1:
name useKVM should be more consistent to kubernetes naming convention

### DIFF
--- a/roles/KubevirtNodeLabeller/defaults/main.yml
+++ b/roles/KubevirtNodeLabeller/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
 # defaults file for KubevirtNodeLabeller
 wait: true
-use_kvm: true
+useKVM: true
 kubevirt_node_labeller_image: "node-labeller"
 kvm_info_nfd_plugin_image: "kvm-info-nfd-plugin"
 kubevirt_cpu_nfd_plugin_image: "cpu-nfd-plugin"
--- a/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
+++ b/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
@@ -46,13 +46,13 @@ spec:
         - name: libvirt
           image: {{ ssp_registry | default('kubevirt') }}/{{ image_name_prefix }}{{ virt_launcher_image }}:{{ virt_launcher_tag }}
           command: ["/bin/sh","-c"]
-{% if use_kvm %}
+{% if useKVM %}
           args: ["libvirtd -d; chmod o+rw /dev/kvm; virsh domcapabilities --machine q35 --arch x86_64 --virttype kvm > /etc/kubernetes/node-feature-discovery/source.d/virsh_domcapabilities.xml; cp -r /usr/share/libvirt/cpu_map /etc/kubernetes/node-feature-discovery/source.d/"]
 {% else %}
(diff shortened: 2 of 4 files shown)


roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
```diff
@@ -46,13 +46,13 @@ spec:
         - name: libvirt
           image: {{ ssp_registry | default('kubevirt') }}/{{ image_name_prefix }}{{ virt_launcher_image }}:{{ virt_launcher_tag }}
           command: ["/bin/sh","-c"]
-{% if use_kvm %}
+{% if useKVM %}
           args: ["libvirtd -d; chmod o+rw /dev/kvm; virsh domcapabilities --machine q35 --arch x86_64 --virttype kvm > /etc/kubernetes/node-feature-discovery/source.d/virsh_domcapabilities.xml; cp -r /usr/share/libvirt/cpu_map /etc/kubernetes/node-feature-discovery/source.d/"]
```
fromanirh: I think in this file we need to support both ways up until the next stable version, included :\

roles/KubevirtNodeLabeller/defaults/main.yml
```diff
@@ -1,7 +1,8 @@
 ---
 # defaults file for KubevirtNodeLabeller
 wait: true
-use_kvm: true
+useKVM: true
+use_kvm: false
```
MarSik: Honestly, I do not like the duplication.

MarSik:
> the stable version 1.0.13 unfortunately included `use_kvm`. We need to preserve backward compatibility one version behind IIRC (for HCO, see [kubevirt/hyperconverged-cluster-operator#276](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/276))

Isn't HCO using the golang API? The name is the same there. The CSV manifests are generated from pieces we provide so we should be ok I think.

fromanirh:
> > the stable version 1.0.13 unfortunately included `use_kvm`. We need to preserve backward compatibility one version behind IIRC (for HCO, see [kubevirt/hyperconverged-cluster-operator#276](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/276))
> 
> Isn't HCO using the golang API? The name is the same there. The CSV manifests are generated from pieces we provide so we should be ok I think.

Yes. Their released code will have golang code which translates the (unchanged) go struct field in the json "use_kvm" key:
```golang
UseKVM  bool   `json:"use_kvm,omitempty"`
```

so our code on the receiving side must handle either "use_kvm" and "useKVM" json keys, unfortunately.

roles/KubevirtNodeLabeller/defaults/main.yml
fromanirh: we have another option:
bury this into the task file, which will set `useKVM=true` also if the `use_kvm` var is true. This will allow us to leave the deployment and the default files unchanged.

fromanirh: We need to check if HCO stable version X can or should consume our operator version X-1. If so, we will need the backward compat. Otherwise, we can happily change only our master and call it done without the ugly backward compat.

fromanirh:
OK, we had a quick chat to the HCO folks and it seems we don't really need the backward compat for the time being. We will re-add later if needed.

PR #110, opened 2019-10-15
node-labeller: use snake-case internally, again.

fromanirh:
When the ansible operator creates the playbook parameters,
it takes all the attributes under `spec:` and creates
variables out of them.
However, the variable names are not taken verbatim, but
they are translated to snake_case:
https://github.com/operator-framework/operator-sdk/blob/v0.11.x/pkg/ansible/runner/runner.go#L259

So the role variables should always use the snake_case variant
automatically created, not the camelCase exposed version.
Out of necessity, this commit undoes most of the changes
done in 83528c8 because we had the following bug:

1. HCO wants to set useKVM=false when running on its CI
2. the ansible operator translates the variable to use_kvm=false
   and pass the parameter to the playbook
3. the playbook runs, and sets the defaults useKVM=true
4. the playbook consumes the useKVM (camelCase) variable, whose
   defaults has been set and never overriden
5. the use_kvm (snake_case) variable, which holds the correct value
   the user set (in the CR spec.useKVM!) is ignored entirely.

So the only fix is expose fields to kubernetes in camelCase (useKVM) but
consume their snake_case counterparts (use_kvm) in the roles.

### DIFF
--- a/roles/KubevirtNodeLabeller/defaults/main.yml
+++ b/roles/KubevirtNodeLabeller/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
 # defaults file for KubevirtNodeLabeller
 wait: true
-useKVM: true
+use_kvm: true
 kubevirt_node_labeller_image: "node-labeller"
 kvm_info_nfd_plugin_image: "kvm-info-nfd-plugin"
 kubevirt_cpu_nfd_plugin_image: "cpu-nfd-plugin"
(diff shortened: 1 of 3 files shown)


ksimon1:
what about https://github.com/MarSik/kubevirt-ssp-operator/blob/master/pkg/apis/kubevirt/v1/types.go#L73?

fromanirh:
> what about https://github.com/MarSik/kubevirt-ssp-operator/blob/master/pkg/apis/kubevirt/v1/types.go#L73?

This is the only correct change. The struct is translated to JSON, and should be `camelCase` because the JSON is still at kubernetes level.

PR #111, opened 2019-10-17
roles: repoinfo: simplify autodetection

fromanirh:
We can do much like kubevirt core does: set the image path
in the manifests themselves inside that introspect at runtime.
This is simpler and much more robust when doing upgrades.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -32,7 +32,7 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
-            - name: IMAGE_NAME
+            - name: IMAGE_REFERENCE
               value: REPLACE_IMAGE
             - name: WATCH_NAMESPACE
               value: ""
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -32,6 +32,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: IMAGE_NAME
+              value: REPLACE_IMAGE
             - name: WATCH_NAMESPACE
               value: ""
             - name: KVM_INFO_TAG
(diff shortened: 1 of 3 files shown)


deploy/operator.yaml
```diff
@@ -32,6 +32,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            - name: IMAGE_NAME
```
MarSik: I would probably use IMAGE_REFERENCE or something (to avoid `name`), but this works as well

roles/KubevirtRepoInfo/tasks/main.yml
```diff
@@ -1,11 +1,8 @@
 ---
 # tasks file for KubevirtRepoInfo
-- name: Extract the operator POD info
-  set_fact:
-    operator_pod: "{{ lookup('k8s', api_version='v1', kind='Pod', label_selector='name=kubevirt-ssp-operator') | from_yaml }}"
 - name: Extract the image name
   set_fact:
-    operator_image_name: "{{ operator_pod['spec']['containers'][0]['image'] }}"
+    operator_image_name: "{{ lookup('env', 'IMAGE_NAME') | default('quay.io/kubevirt/kubevirt-ssp-operator-container:latest', true }}"
```
MarSik: Aren't you missing a closing parenthesis?

fromanirh: I do!

deploy/operator.yaml
fromanirh: since I have to reupload, I'll change this.

fromanirh:
ok, seems to work as expected

fromanirh:
seems to work ok with HCO (manual installation using `deploy.sh`) as well

fromanirh:
closed by mistake (again)

PR #122, opened 2019-11-11
updated common templates to version v0.8.0

MarSik:
updated common templates to version v0.8.0


MarSik:
One check reached timeout (test 08), but essentially the same thing for templates (10) took more time and succeeded at the end. We need to relax and unify the timeouts, but the result was good enough for this PR.

PR #123, opened 2019-11-11
functests: validator: add troubleshooting helpers

fromanirh:
We need to learn why the functest#09 fails so often.


MarSik:
I guess the simple timeout is too low now...

fromanirh:
> I guess the simple timeout is too low now...

Which is weird: I only increased timeouts, even these for simple tests!

PR #124, opened 2019-11-11
Commit generated manifests for HCO

MarSik:
We have an issue with permissions. Apparently the csv-generator does not see the updated permissions until we run make manifests and commit the result into the repository. I do not like that much, but the alternative is downloading the full operator-sdk into the Docker image to be able to run make manifests from there.


MarSik:
The usual timeout we are trying to solve in #123.

PR #125, opened 2019-11-12
Hotfix csv generator to support inline yaml

MarSik:
The csv-generator was originally written to replace
env values in the form of

- name: NAME
  value: ....

But the currently generated manifest uses the inline form:

- {name: NAME, value: ...}

This patch updates the script so both forms are supported.

Signed-off-by: Martin Sivak <msivak@redhat.com>

### DIFF
--- a/build/csv-generator.sh
+++ b/build/csv-generator.sh
@@ -15,7 +15,9 @@ replace_env_var() {
     local value_offset="                  "
     local key=$1
     local var=$2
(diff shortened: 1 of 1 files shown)


MarSik:
The issue was discovered when the last commit with committed manifest was used during container build: https://github.com/MarSik/kubevirt-ssp-operator/commit/2baa7aa75b347d042a43e9aa30a8e02bd3c14918#diff-0e83c1a6816c2de91b770e426af107b2R110

PR #126, opened 2019-11-12
Install pip3 packages in the users home folder

rmohr:
Avoid the need for extra privileges when running `make manifests`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ operator-sdk:
 	chmod 0755 operator-sdk
 
 operator-courier:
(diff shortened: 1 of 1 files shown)


rmohr:
@MarSik maybe not using `--user` was intentional. Just let me know.

MarSik:
@ksimon1 Karel? Was there a reason or is this fine? I am ok with that change myself.

ksimon1:
@MarSik no, there was no reason. Change looks ok.

PR #136, opened 2019-12-09
[WIP ]decreased timeout time in "wait" tasks

ksimon1:
decreased timeout time in "wait" tasks
This change should prevent uncontrollable "freeze" of ssp operator.
This "freeze" happens in this scenario(mostly during tests): e.g. template-validator
is deployed. Then ansible moves to "Wait for the template-validator to start" task.
But when user (or script) deletes template-validator CR, operator keeps trying to test
if validator is online or not (in this case for more than 3000 seconds) and then operator fails
and whole container is restarted. This patch decrease this time to only 60 seconds and prevents restarts
of container by adding flag ignore_errors to "wait" tasks (for now only validator + node-labeller have this task)

Signed-off-by: Karel Šimon <ksimon@redhat.com>

### DIFF
--- a/roles/KubevirtNodeLabeller/tasks/main.yml
+++ b/roles/KubevirtNodeLabeller/tasks/main.yml
@@ -49,8 +49,9 @@
     namespace: "{{ nl.result.metadata.namespace }}"
   register: nl_status
   delay: 10
-  retries: 300
+  retries: 6
   until: nl_status.resources[0].status.currentNumberScheduled == nl_status.resources[0].status.numberReady | default(false)
+  ignore_errors: yes
 
 - name: Set available condition
   k8s_status:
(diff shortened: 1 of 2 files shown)


fromanirh:
My concern is: how does this affect real, production deployments? Can enable this just in tests, for starters? Would it make sense to have a different wait time for tests and for production?

ksimon1:
I agree, for production deployments we can stay with 300 iterations, but for tests we need e.g. only 2-3 iterations. Otherwise tests will fail, even though they should not.
Another question is with ignore_errors attribute. If the case i mentioned happens, the container will fail, and it will be restarted. And i don't think it should happen.

petrkotas:
I agree with @ksimon1 in tests we aim for speed. In production, I have no opinion as I have not seen it run there.

PR #137, opened 2019-12-10
[WIP] timeout in "wait" tasks can be set via special variable

ksimon1:
add ignore_errors to "wait" tasks which prevents restarts of container

This change should prevent uncontrollable "freeze" of ssp operator.
This "freeze" happens in this scenario(mostly during tests): e.g. template-validator
is deployed. Then ansible moves to "Wait for the template-validator to start" task.
But when user (or script) deletes template-validator CR, operator keeps trying to test
if validator is online or not (in this case for more than 3000 seconds) and then operator fails
and whole container is restarted. This patch decrease this time to only 40 seconds during tests and prevents restarts
of container by adding flag ignore_errors to "wait" tasks (for now only validator + node-labeller have this task)

I accidentally closed original PR https://github.com/MarSik/kubevirt-ssp-operator/pull/136

/cc @petrkotas, @fromanirh, @MarSik

Signed-off-by: Karel Šimon <ksimon@redhat.com>


pkg/apis/kubevirt/v1/types.go
```diff
@@ -69,13 +69,15 @@ type VersionSpec struct {
...
-	UseKVM  bool   `json:"useKVM"`
+	Version                 string `json:"version,omitempty"`
+	UseKVM                  bool   `json:"useKVM"`
+	NumberOfLabellerRetries int    `json:"numberOfLabellerRetries,omitempty"`
 }
 
 type TemplateValidatorSpec struct {
 	Version                   string `json:"version,omitempty"`
 	TemplateValidatorReplicas int    `json:"templateValidatorReplicas,omitempty"`
+	NumberOfValidatorRetries  int    `json:"numberOfLabellerRetries,omitempty"`
```
fromanirh: Typo: should be `numberOfValidatorRetries`.

I'd actually use `MaxRetries` or `MaxRetriesNumber` for both structs.

PR #144, opened 2020-02-12
updated api group in crd and crs to ssp.kubevirt.io

ksimon1:
The reason for this is to avoid that there are API clashes in the Kube API, where one operator is overriding the APIs of other operators.


orenc1:
```
The CustomResourceDefinition "kubevirtcommontemplatesbundles.kubevirt.io" is invalid: metadata.name: Invalid value: "kubevirtcommontemplatesbundles.kubevirt.io": must be spec.names.plural+"."+spec.group
```

PR #154, opened 2020-03-13
Fix the paused annotation docs

phoracek:
ssp prefix was dropped


MarSik:
The patch updates just the README, there is no reason to wait for CI.

PR #156, opened 2020-03-24
Removed node-labeller from ssp-operator

vatsalparekh:
Signed-off-by: Vatsal Parekh <vparekh@redhat.com>


ksimon1:
The PR looks good. We just need to wait until node-labeller is moved into kubevirt

ksimon1:
This needs to be removed as well https://github.com/MarSik/kubevirt-ssp-operator/blob/master/hack/install-operator.sh#L12

PR #166, opened 2020-05-21
Removed remaining node-labeller files

vatsalparekh:
Remove the files missed in #156 
Signed-off-by: Vatsal Parekh <vparekh@redhat.com>


vatsalparekh:
@ksimon1

PR #167, opened 2020-05-21
remove labeller from manifests and csv generator

ksimon1:
labeller is now part of kubevirt core.


vatsalparekh:
approve

PR #170, opened 2020-05-25
revert removing of labeller

ksimon1:
Due to issues with olm upgrade, labeller can't be deleted from ssp yet.

//cc @tiraboschi


roles/KubevirtNodeLabeller/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# vars file for KubevirtSsp
```
orenc1, line 2: ?

PR #171, opened 2020-05-25
Add container that watches cert changes and updates validator webhook

akrejcir:
The definition of template validator webhook - ValidatingWebhookConfiguration
has a 'CaBundle' field which is used to check the certificate of the webhook
server.

This patch adds a container to the pod that updates the 'CaBundle' filed
when a new CA file is detected.

Signed-off-by: Andrej Krejcir <akrejcir@redhat.com>


akrejcir:
@ksimon1 @omeryahud, can you take a look, if this approach is ok? It is still a draft, there are a few TODOs, that I will add.

internal/webhook-updater/app.go
```diff
@@ -0,0 +1,257 @@
...
+	quitSignal := registerSignalHandlers()
+	err = app.processFileChanges(client, quitSignal)
+	if err != nil {
+		Log.Errorf("Error watching for certificates: %s", err)
+		os.Exit(1)
+	}
+}
+
+func (app *App) parseAndValidateFlags() error {
+	flag.StringVarP(&app.kubeconfig, "kubeconfig", "c", "", "absolute path to the kuebconfig file")
```
ksimon1, line 61: typo `kuebconfig` - `kubeconfig`

internal/webhook-updater/app.go
```diff
@@ -0,0 +1,257 @@
...
+				}
+				Log.Errorf("An error occurred when watching %s. Reason: %s", directory, err)
+			}
+		}
+	}()
+
+	return filesChanged, watcher, nil
+}
+
+func notify(channel chan struct{}) {
```
ksimon1, line 186: Is this function missing something?

omeryahud: This function makes sure to send to the channel only if its buffer is empty, in order to not block the calling routine if the refresh was not performed yet (if the channel is not empty, sending another signal is pointless because whatever caused the second signal, will be applied once the channel is being read from)

internal/webhook-updater/app.go
```diff
@@ -0,0 +1,257 @@
...
+
+	filesChanged := make(chan struct{}, 1)
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				if event.Op != fsnotify.Chmod {
```
omeryahud, line 171: In an event of `fsnotify.Remove`, the process would continuously fail until a new cert file is `fsnotify.Create`d.
Wouldn't it be a good idea to also ignore the `fsnotify.Remove` event and handle the update in the `fsnotify.Create` event that might follow? Because the `caBundle` would not be updated because of `certBytes, err := ioutil.ReadFile(fileName)` in `refreshWebhook`

internal/webhook-updater/app.go
akrejcir: Done

internal/webhook-updater/app.go
akrejcir: Done

PR #181, opened 2020-06-08
fix labeller when no kvm device is available

ksimon1:
When node doesn't have access to kvm, labeller stays in pending state. This commit
updates kvm requirements and if kvm is not available, labeller starts, but doesn't
label the node.
//cc @tiraboschi, @omeryahud 


functests/06-test-deploy-node-labeller.sh
```diff
@@ -14,26 +14,6 @@ oc apply -n ${TEST_NS} -f "${SCRIPTPATH}/node-labeller-unversioned-cr.yaml" || e
 
 wait_node_labeller_running ${TEST_NS} 5 60
 
-for idx in $( seq 1 30); do
```
omeryahud, line 17: Why was this removed? It would be good to keep this for nodes that do have kvm

ksimon1: Because Travis doesn't have access to KVM, so it will not add any labels

omeryahud: Functional tests are also run on the developer's machine, maybe it would be a good idea to modifiy the test to look for labels on nodes that do have access to KVM

PR #186, opened 2020-06-11
Implement upgrade path for SSP operator

omeryahud:
An upgrade of the SSP operator from versions 1.0.22 to a higher version would cause duplicate CRDs due to a change to the CRDs api group:
```
$ oc get crds | grep "aggre\|labeller\|validator\|common"

kubevirtcommontemplatesbundles.kubevirt.io                       2020-06-03T14:07:02Z
kubevirtcommontemplatesbundles.ssp.kubevirt.io                   2020-06-04T10:26:17Z
kubevirtmetricsaggregations.kubevirt.io                          2020-06-03T14:07:02Z
kubevirtmetricsaggregations.ssp.kubevirt.io                      2020-06-04T10:26:17Z
kubevirtnodelabellerbundles.kubevirt.io                          2020-06-03T14:07:02Z
kubevirtnodelabellerbundles.ssp.kubevirt.io                      2020-06-04T10:26:17Z
kubevirttemplatevalidators.kubevirt.io                           2020-06-03T14:07:02Z
kubevirttemplatevalidators.ssp.kubevirt.io                       2020-06-04T10:26:17Z
```
This PR is introducing the following changes:
- The operator now manages the CRs in the `ssp.kubevirt.io` group only.
- When the operator is deployed and reconciling starts, if any resources exist from a previous installation, the operator will take ownership of them. For example, `KubevirtTemplateValidator.ssp.kubevirt.io` will take ownership of all resources created by `KubevirtTemplateValidator.kubevirt.io`, if it existed
- The following `status` fields are added to the CR to indicate a successful upgrade:
  - ```
    status:
      operatorVersion: "v1.0.36"
      targetVersion: "v1.0.36"
      observedVersion: ""
    ```

  * `operatorVersion` - States the version of the deployed operator
  * `targetVersion` - States the desired version of deployed resources
  * `observedVersion` - States the version of currently deployed resources

An upgrade is complete when the `Available` condition is `True`, and when `targetVersion` == `observedVersion`.

### DIFF
--- a/roles/KubevirtCommonTemplatesBundle/tasks/main.yml
+++ b/roles/KubevirtCommonTemplatesBundle/tasks/main.yml
@@ -1,46 +1,113 @@
 ---
+- name: Set operatorVersion and targetVersion
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
+    status:
+      operatorVersion: "{{ operator_version }}"
+      targetVersion: "{{ operator_version }}"
+
 # tasks file for KubevirtCommonTemplatesBundle
 - name: Install VM templates
   k8s:
     state: present
-    namespace: "{{ meta.namespace }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
     definition: "{{ item | from_yaml }}"
+    apply: yes
   with_items: "{{ lookup('file', 'common-templates-'+ version +'.yaml').split('\n---\n') | select('search', '(^|\n)[^#]') | list }}"
   register: ct_status
+
+# Get all templates
+- name: Fetching all templates
+  set_fact:
+    templates: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template') }}"
+
+- name: Fetch old cr
+  block:
+  - set_fact:
+      old_cr: "{{ lookup('k8s', api_version='kubevirt.io/v1', kind='KubevirtCommonTemplatesBundle') }}"
+  - set_fact:
+      old_cr_exists: true
+  rescue:
+  - set_fact:
+      old_cr_exists: false
+
+- name: Filter for templates owned by the old cr
+  set_fact:
+    old_cr_templates: "{{ templates | k8s_owned_by(old_cr) }}"
+  when: "{{ old_cr_exists==true }}"
+
+# Inject ownerReferences
+- name: Inject owner references for KubevirtCommonTemplatesBundle
+  include_role: 
+    name: ClaimOwnership
+  vars:
+    object: "{{ item }}"
+    owner: "{{ cr_info }}"
+  when: "{{ old_cr_exists==true }}"
+  with_items: "{{ old_cr_templates }}" # Templates
 
 - name: "Count all new templates in file"
   set_fact:
     new_templates: "{{ lookup('file', 'common-templates-'+ version +'.yaml').split('\n---\n') | select('search', '(^|\n)[^#]') | list | length }}"
 
-
 - name: "Set label"
   set_fact:
     label: "template.kubevirt.io/version={{ version }}"
-  
+
 - name: "Get all templates"
   set_fact:
-    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template', label_selector=label)|length }}"
+    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template', label_selector=label) | length }}"
+
+- name: "Set Available status"
+  set_fact:
+    available: "{{ true if new_templates <= deployed_templates_after else false }}"
 
 - name: Set progressing condition
   operator_sdk.util.k8s_status:
-    api_version: ssp.kubevirt.io/v1
-    kind: KubevirtCommonTemplatesBundle
-    name: "{{ meta.name }}"
-    namespace: "{{ meta.namespace }}"
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
     conditions:
     - type: Progressing
-      status: "{{ 'True' if new_templates > deployed_templates_after else 'False' }}"
+      status: "False"
       reason: "progressing"
-      message: "Templates progressing."
+      message: "Templates progressing (deployed templates: {{ deployed_templates_after }}, desired deployed templated: {{ new_templates }})."
 
 - name: Set available condition
   operator_sdk.util.k8s_status:
-    api_version: ssp.kubevirt.io/v1
-    kind: KubevirtCommonTemplatesBundle
-    name: "{{ meta.name }}"
-    namespace: "{{ meta.namespace }}"
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
     conditions:
     - type: Available
-      status: "{{ 'True' if new_templates <= deployed_templates_after else 'False' }}"
+      status: "True"
       reason: "available"
-      message: "Common templates available."
+      message: "Common templates available (deployed templates: {{ deployed_templates_after }}, desired deployed templated: {{ new_templates }})."
+
+- name: Set degraded condition
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
+    conditions:
+    - type: Degraded
+      status: "False"
+      reason: "degraded"
+      message: "Templates degraded (deployed templates: {{ deployed_templates_after }}, desired deployed templated: {{ new_templates }})."
+
+# Update observerVersion when the CR is Available to indicate a successfull upgrade
+- name: Set observedVersion
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
+    status:
+      observedVersion: "{{ operator_version }}"
+  when: "{{ available==true }}"
--- a/roles/KubevirtMetricsAggregation/tasks/main.yml
+++ b/roles/KubevirtMetricsAggregation/tasks/main.yml
@@ -1,9 +1,75 @@
 ---
+- name: Set operatorVersion and targetVersion
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
+    status:
+      operatorVersion: "{{ operator_version }}"
+      targetVersion: "{{ operator_version }}"
+
 # tasks file for KubevirtMetricsAggregation
 - name: Install VMI count aggregation rule
   k8s:
     state: present
     namespace: "{{ meta.namespace }}"
     definition: "{{ item | from_yaml }}"
+    apply: yes
   with_items: "{{ lookup('template', 'aggregation-rule-vmi-count.yaml.j2').split('\n---\n') | select('search', '(^|\n)[^#]') |list }}"
+  register: promrules
 
+- name: Inject owner references for KubevirtMetricsAggregation
+  include_role:
+    name: ClaimOwnership
+  vars:
+    object: "{{ item.result }}"
+    owner: "{{ cr_info }}"
+  with_items: "{{ promrules.results }}"
+
+- name: Set available condition
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
+    name: "{{ cr_info.metadata.name }}"
+    namespace: "{{ cr_info.metadata.namespace }}"
+    conditions:
+    - type: Available
+      status: "True"
+      reason: "available"
+      message: "KubevirtMetricsAggregation is available."
+
+# There is no logic for the Progressing/Degraded conditions for KubevirtMetricsAggregation
+# so using these constant conditions
+- name: Set progressing condition
+  operator_sdk.util.k8s_status:
+    api_version: "{{ cr_info.apiVersion }}"
+    kind: "{{ cr_info.kind }}"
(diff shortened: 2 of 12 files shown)


roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -8,6 +8,28 @@
   with_items: "{{ lookup('file', 'common-templates-'+ version +'.yaml').split('\n---\n') | select('search', '(^|\n)[^#]') | list }}"
   register: ct_status
 
+# Get all templates (including templates deployed by a previous version)
+- name: Fetching all templates to inject ownerReference
+  set_fact:
+    templates: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template', label_selector='template.kubevirt.io/type=base') }}"
```
MarSik, line 14: This will also claim ownership of all user created base templates! You need to be more selective and only claim templates created by the operator.

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -8,6 +8,28 @@
...
+
+# Inject ownerReferences to all templates
+- name: Inject ownerRef into the Templates definition
```
MarSik, line 17: I suggest you create a new role - ClaimOwnership - and invoke it from here (and from the other places). Just to avoid repeating this block so much.

```
- include_role:
    name: ClaimOwnership
    vars:
      object: "{{ item }}"
      owner: "{{ cr_info }}" -- and you might be able to omit this one if you use cr_info everywhere consistently
```

roles/KubevirtMetricsAggregation/tasks/main.yml
```diff
@@ -6,4 +6,21 @@
     namespace: "{{ meta.namespace }}"
     definition: "{{ item | from_yaml }}"
   with_items: "{{ lookup('template', 'aggregation-rule-vmi-count.yaml.j2').split('\n---\n') | select('search', '(^|\n)[^#]') |list }}"
+  register: promrules
 
+- name: Inject ownerRef into the PrometheusRule definition
```
MarSik, line 11: This is not needed I think. You are not looking up existing PrometheusRoles and the ansible operator injects the proper reference automatically to newly created ones.

roles/KubevirtNodeLabeller/tasks/main.yml
```diff
@@ -4,13 +4,55 @@
...
   with_items: "{{ lookup('template', 'kubevirt-node-labeller-roles.yaml.j2').split('\n---\n') | select('search', '(^|\n)[^#]') | list }}"
-
+  register: roles
 - name: Create the node labeller daemon set
   k8s:
     state: present
     definition: "{{ lookup('template', 'kubevirt-node-labeller-ds.yaml.j2') | from_yaml }}"
   register: nl
-
+# Actively inject owner references in order to adopt existing resources during an upgrade
```
MarSik, line 13: Same here, are you sure you need this? You are reading the objects from file and then posting them. The ansible operator sdk should do the right thing automatically.

roles/KubevirtTemplateValidator/tasks/main.yml
```diff
@@ -10,6 +10,49 @@
     definition: "{{ item | from_yaml }}"
   with_items: "{{ lookup('template', 'service.yaml.j2').split('\n---\n') | select('search', '(^|\n)[^#]') | list }}"
   register: tv
+# Actively inject owner references in order to adopt existing resources during an upgrade
```
MarSik, line 13: And same here..

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
MarSik: You can be more fancy even and create a module too. It is just a Python file..

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Notice the label selector `template.kubevirt.io/type=base`, do user created templates have this label as well?

omeryahud: I assume they don't, are you worried about templates cloned by the user?

roles/KubevirtMetricsAggregation/tasks/main.yml
omeryahud: The operator does not inject the reference if the object already exists, so I need to do it manually here in case the PrometheusRule already exists

roles/KubevirtNodeLabeller/tasks/main.yml
omeryahud: Again here, if the DS already exists from a previous deployment, the owner reference would not be set

roles/KubevirtTemplateValidator/tasks/main.yml
omeryahud: Same here

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
MarSik: `template.kubevirt.io/type=base` says nothing about ownership, it marks the template as a base template (as opposed to a template in general or template created from existing VM). The UI only displays templates with type=base. Including user defined ones.

roles/KubevirtMetricsAggregation/tasks/main.yml
MarSik: What about if it changes? This happens during upgrade so in theory the content might be different.

MarSik: I wonder if using `apply: yes` in the k8s task would fix that cleanly: https://docs.ansible.com/ansible/latest/modules/k8s_module.html#parameter-apply

It was introduced to Ansible 2.9 though, so you need to check what version the operator uses.

omeryahud: According to this, owner references will never be automatically injected to existing objects:
https://github.com/operator-framework/operator-sdk/blob/master/website/content/en/docs/ansible/reference/retroactively-owned-resources.md

omeryahud: So if it changes, the resource would be updated, but no new ownerRef would be set

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Will try creating a new role, thanks!

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Do you have any suggestions on how to query for operator created templates?

MarSik: Take the list you got from the query and filter it based on the expected old owner reference?

omeryahud: I'll try that, thanks

roles/KubevirtMetricsAggregation/tasks/main.yml
MarSik: I see.

roles/KubevirtNodeLabeller/tasks/main.yml
MarSik: What I do not like here is that you list all the different types you want to own. That will make maintenance hard if the source yaml file changes. Can you just iterate over all the created objects instead?

roles/KubevirtTemplateValidator/tasks/main.yml
MarSik: Same objection to the duplication based on kind. In fact, the `results[0]` based indexing assumes a lot about the source file.

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Done

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Done

roles/KubevirtNodeLabeller/tasks/main.yml
omeryahud: Could you take another look please? I've run the role for each resource that needs to be owned by the CR

roles/KubevirtTemplateValidator/tasks/main.yml
omeryahud: I don't see any other way to do it other than fetching the resources I want to own from the cluster one by one and patching them

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -41,6 +84,23 @@
...
+    namespace: "{{ meta.namespace }}"
+    conditions:
+    - type: Degraded
+      status: "{{ 'True' if degraded else 'False' }}"
+      reason: "degraded"
+      message: "Templates degraded (deployed templates: {{ deployed_templates_after }}, desired deployed templated: {{ new_templates }})."
+
+- name: Reque if KubevirtCommonTemplatesBundle not available
+  fail:
+    msg: "Requeing until KubevirtTemplateValidator is available"
```
MarSik, line 105: until all templates are available?

But that can never happen in fact, templates are not "active", so once you get here you have them all.

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -41,6 +84,23 @@
     namespace: "{{ meta.namespace }}"
     conditions:
     - type: Available
-      status: "{{ 'True' if new_templates <= deployed_templates_after else 'False' }}"
+      status: "{{ 'True' if available else 'False' }}"
       reason: "available"
-      message: "Common templates available."
+      message: "Common templates available (deployed templates: {{ deployed_templates_after }}, desired deployed templated: {{ new_templates }})."
+
+- name: Set degraded condition
```
MarSik, line 91: Same here, how can this happen? The playbook installs all of them directly..

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -5,21 +5,64 @@
...
+      old_cr: "{{ lookup('k8s', api_version='kubevirt.io/v1', kind='KubevirtCommonTemplatesBundle') }}"
+  - set_fact:
+      old_cr_exists: true
+  rescue:
+  - set_fact:
+      old_cr_exists: false
+
+- name: Filter for templates owned by the old cr
+  set_fact:
+    old_cr_templates: "{{ (old_cr_templates | default([])) + ([item] if (item.metadata.ownerReferences is defined and item.metadata.ownerReferences[0].uid==old_cr.metadata.uid) else []) }}"
```
MarSik, line 29: I do not see the filtering here.. It is also pretty unreadable. I suggest writing a python filter and then using for example:

`{{ old_cr_templates |k8s_owned_by(old_cr) }}`

A example of how to do that:
- create `filter_plugins` dir in the role (as sibling to tasks dir)
- in there create a file `k8s.py`

And in the file write something like:

```
def k8s_owned_by(objects, owner):
    return [o for o in objects if ....]

class FilterModule(object):
    '''Extra filters...'''
    def filters(self):
        return {
            'k8s_owned_by': k8s_owned_by
        }
```

roles/KubevirtTemplateValidator/tasks/main.yml
MarSik: The issue with tv.results[0].result is that the code will break when someone changes the order of objects in service.yaml.j2. Why not simply iterate over all of them?

roles/KubevirtNodeLabeller/tasks/main.yml
```diff
@@ -3,20 +3,47 @@
...
 
+# Actively inject owner references in order to adopt existing resources during an upgrade
+- name: Inject owner references for KubevirtNodeLabellerBundle
+  include_role: 
+    name: ClaimOwnership
+  vars:
+    object: "{{ item }}"
+    owner: "{{ cr_info }}"
+  with_list:
+  - "{{ roles.results[0].result }}" # ServiceAccount
```
MarSik, line 25: The issue with roles.results[0].result is that the code will break when someone changes the order of objects in kubevirt-node-labeller-ds.yaml.j2. Why not simply iterate over all of them?

omeryahud: The issue is that the file kubevirt-node-labeller-roles.yaml.j2 contains cluster-scoped resources that cannot be owned by a namespaced object

roles/KubevirtTemplateValidator/tasks/main.yml
omeryahud: The issue is that the file service.yaml.j2 contains cluster-scoped resources that cannot be owned by a namespaced object (ClusterRole, ClusterRoleBinding)

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: Correct, will remove this

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: I set this condition in favor of HCO, do you think it would be better to place a constant value for the status? Because you are correct, once the playbook reaches this point all templates should exist already and the condition status would never be true

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
omeryahud: This task iterates over all templates fetched by the `Fetching all templates` task, and appends all templates that have the `old_cr` as their owner to `old_cr_templates`

roles/KubevirtNodeLabeller/tasks/main.yml
MarSik: Can you filter out those algorithmically?

omeryahud: Only way to filter those out is by name/kind

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
MarSik: Yes, use a constant. I think it even was like that in the past.

omeryahud: Done

akrejcir:
Can you squash related commits together, so that each commit makes sense on its own? It would be easier to review and look better in the history.

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -1,46 +1,121 @@
...
   set_fact:
     label: "template.kubevirt.io/version={{ version }}"
-  
+
 - name: "Get all templates"
   set_fact:
-    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template', label_selector=label)|length }}"
+    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.results[0].result.apiVersion, kind='template', label_selector=label) | length }}"
+
+- name: "Set Progressing status"
```
MarSik, line 64: You do not need the facts here anymore as you do not use them

PR #202, opened 2020-07-17
Changed way how templates are deployed.

ksimon1:
Previously templates were deployed one by one. This causes issues for
HCO operator (timeout when setting available condition) and HCO operator was deleted by kubernetes. This commit updates the deploy task to deploy all templates at once


roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -50,14 +49,12 @@
...
-    objectKind: "templates"
-    owner: "{{ cr_info }}"
-  when: "{{ old_cr_exists==true }}"
-  with_items: "{{ old_cr_templates }}" # Templates
+  k8s:
+    state: present
+    namespace: "{{ cr_info.metadata.namespace }}"
+    merge_type: ['merge', 'json']
+    resource_definition: "{{ old_cr_templates | k8s_inject_ownership(cr_info) }}"
+  when: "{{ old_cr_templates | length > 0 }}"
```
omeryahud, line 57: I looked at travis logs and it fails because `old_cr_templates` is not defined, probably because `old_cr_exists = false`, I think `(old_cr_templates is defined) and (old_cr_templates | length > 0)` might solve this

roles/KubevirtCommonTemplatesBundle/filter_plugins/k8s_inject_ownership.py
```diff
@@ -0,0 +1,26 @@
...
+    owner_ref = {
+        "apiVersion": owner["apiVersion"],
+        "kind": owner["kind"],
+        "name": owner['metadata']["name"],
+        "uid": owner['metadata']['uid']
+    }
+
+    ownerUID = owner['metadata']['uid']
+
+    for obj in objects:
```
omeryahud, line 19: This part seems a bit odd to me, I don't see how it adds `owner_ref` to a template (`object`) that doesn't already have it

omeryahud:
I just noticed the PR is still WIP, feel free to ignore my premature comments

roles/KubevirtCommonTemplatesBundle/filter_plugins/k8s_inject_ownership.py
```diff
@@ -0,0 +1,32 @@
...
+        if obj['metadata']['ownerReferences'] is not None:
+            found = False
+            for index, ref in obj['metadata']['ownerReferences']:
+                if ref['uid'] == ownerUID:
+                    obj["metadata"]["ownerReferences"][index] = owner_ref
+                    found = True
```
omeryahud, line 25: You can break out of the loop once the ref is found

omeryahud:
@ksimon1 Have you noticed improvements in deployment/upgrade speed?

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -50,14 +49,12 @@
...
 - name: Inject owner references for KubevirtCommonTemplatesBundle
-  include_role: 
-    name: ClaimOwnership
-  vars:
-    object: "{{ item }}"
-    objectKind: "templates"
-    owner: "{{ cr_info }}"
-  when: "{{ old_cr_exists==true }}"
-  with_items: "{{ old_cr_templates }}" # Templates
+  k8s:
```
omeryahud, line 52: The `ClaimOwnership` role can be removed as it is not being used anymore

ksimon1: And what about here https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/roles/KubevirtNodeLabeller/tasks/main.yml#L30?

omeryahud: Ah, I was too focused on common-templates.
Maybe you can replace it with your new filter?

roles/KubevirtNodeLabeller/tasks/main.yml
```diff
@@ -24,17 +24,23 @@
     merge_type: ['merge', 'json']
   register: nl
 
+- name: "collect all objects"
```
omeryahud, line 27: I think this looks more readable:
```
- name: "Collect all objects"
  set_fact:
    objects:
    - "{{ roles.results[0].result }}" # ServiceAccount
    - "{{ roles.results[4].result }}" # ConfigMap
    - "{{ nl.result }}"
```

PR #203, opened 2020-07-22
Updated references and import paths to the new repository location

omeryahud:


omeryahud:
@ksimon1 @akrejcir Can you take a look please?

build/generate.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 ./vendor/k8s.io/code-generator/generate-groups.sh \
 	all \
-	github.com/MarSik/kubevirt-ssp-operator/pkg/client \
-	github.com/MarSik/kubevirt-ssp-operator/pkg/apis \
+	github.com/kubevirt/kubevirt-ssp-operator/pkg/client \
```
akrejcir, line 4: Can we change the package path to `kubevirt.io/kubevirt-ssp-operator/...`, like other kubevirt components have?

akrejcir: I see that some components use the github path. So this is probably ok.

omeryahud: Yea, I noticed all other components except for kubevirt are prefixed with `github.com` and not `kubevirt.io`, maybe we can change that in the future

PR #223, opened 2020-08-26
[OCP CI] Deploying the operator image on OCP CI using tags instead of digest

omeryahud:
Deploying the operator image on OCP CI using tags instead of digest because the test container in OCP ci does not support docker-in-docker so cannot fetch the image digest before deploying

Fixes OCP CI deployment

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Release note**:

```release-note
NONE
```


omeryahud:
/cc @ksimon1 @akrejcir 
This is a small fix to allow OCP CI rehearsals to pass here https://github.com/openshift/release/pull/11173
and then we can activate OCP CI presubmits

PR #226, opened 2020-08-30
Added placement API to TemplateValidator

omeryahud:
The placement of Template Validator pods can be controlled using the following spec fields:
```
spec.nodeSelector
spec.affinity
spec.tolerations
```

This PR adds the placement API as defined here: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/718
to Template Validator

/cc @tiraboschi 
Template Validator should only accept placement parameters from `HCO_CR.spec.infra`

**Release note**:

```release-note
It is now possible to define custom placement for KubevirtTemplateValidator using native Affinities, Node Selectors and Tolerations
```

### DIFF
--- a/functests/10-test-template-validator-placement.sh
+++ b/functests/10-test-template-validator-placement.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+SCRIPTPATH=$( dirname $(readlink -f $0) )
+source ${SCRIPTPATH}/testlib.sh
+
+RET=0
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/10-template-validator-unversioned-cr.yaml" || exit 2
+
+# Wait for the operator to create the deployment, we don't care if pods are actually ready, as
+# we only check for the pod scheduling fields
+DEPLOYMENT_FOUND=false
+for i in {1..20}; do
+  oc get -n ${TEST_NS} deploy virt-template-validator
+  EXIT_CODE=$?
+  if (( $EXIT_CODE == 0 )); then
+    DEPLOYMENT_FOUND=true
+    break
+  else
+    sleep 10
+  fi
+done
+
+if [ "$DEPLOYMENT_FOUND" == "false" ]; then
+  echo "virt-template-validator deployment was not found"
+  exit 1
+fi
+
+NODE_SELECTOR=$(oc get -n ${TEST_NS} deploy virt-template-validator -ojson | jq '.spec.template.spec.nodeSelector.testKey' | tr -d '"')
+if [ "$NODE_SELECTOR" != "testValue" ]; then
+  echo $NODE_SELECTOR
+  echo "template validator deployment is missing proper nodeSelector"
+  RET=1
+fi
+
+TOLERATION=$(oc get -n ${TEST_NS} deploy virt-template-validator -ojson | jq '.spec.template.spec.tolerations[0].key' | tr -d '"')
+if [ "$TOLERATION" != "testKey" ]; then
+  echo $TOLERATION
+  echo "template validator deployment is missing proper tolerations"
+  RET=1
+fi
+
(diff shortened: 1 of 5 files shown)


omeryahud:
@shwetaap Hey Shweta, this should give you a hint on how to add the placement API to node-labeller
/cc @ksimon1 @akrejcir @vatsalparekh

roles/KubevirtTemplateValidator/templates/service.yaml.j2
```diff
@@ -58,6 +58,9 @@ spec:
       labels:
         kubevirt.io: virt-template-validator
     spec:
+      affinity: {{ cr_info['spec']['affinity'] | default("", true) | from_yaml }}
+      nodeSelector: {{ cr_info['spec']['nodeSelector'] | default("", true) | from_yaml }}
```
akrejcir, line 62: This looks like assigning of incompatible types, does the code work?

The `nodeSelector` field of `core.v1.PodSpec` struct has type `map[string]string`, but the `nodeSelector` in `TemplateValidatorSpec` is `v1.NodeSelector`.

akrejcir (changes requested):

roles/KubevirtTemplateValidator/templates/service.yaml.j2
omeryahud: The code works, but I will take another look at the types maybe I missed something

omeryahud: You are correct, I will change that

functests/10-test-template-validator-placement.sh
```diff
@@ -0,0 +1,38 @@
...
+source ${SCRIPTPATH}/testlib.sh
+
+RET=0
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/10-template-validator-unversioned-cr.yaml" || exit 2
+
+# Wait for the operator to create the deployment, we don't care if pods are actually ready, as
+# we only check for the pod scheduling fields
+sleep 100
```
akrejcir, line 13: I think it may be better to call `oc get` in a loop with a timeout, instead of hard-coded `sleep`. But I don't know how fast the deployment is, or if it would save much time.

akrejcir (changes requested):
Just a small comment. Can you also squash the commits?

functests/10-test-template-validator-placement.sh
omeryahud: I agree, will change taht

omeryahud: @akrejcir done

functests/10-test-template-validator-placement.sh
```diff
@@ -0,0 +1,53 @@
...
+RET=0
+TEST_NS="${KV_NAMESPACE}"
+
+oc create -n ${TEST_NS} -f "${SCRIPTPATH}/10-template-validator-unversioned-cr.yaml" || exit 2
+
+# Wait for the operator to create the deployment, we don't care if pods are actually ready, as
+# we only check for the pod scheduling fields
+DEPLOYMENT_FOUND=false
+for i in {1..20}; do
+  oc get -n ${TEST_NS} deploy virt-template-validator
```
akrejcir, line 15: Do we want to see the `oc get` output, or redirect it to `/dev/null`?
I think we should redirect, but it may be useful for debugging.

akrejcir (commented):
Another small comment.

functests/10-test-template-validator-placement.sh
omeryahud: when running `make functests` the actual test files are wrapped with `test-runner.sh` which encapsulates `stdout` unless a test fails, see here: https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/functests/test-runner.sh#L37

akrejcir: Ok, I didn't know about that.

PR #229, opened 2020-09-04
Add Placement API for NodeLabeller

shwetaap:
Signed-off-by: Shweta Padubidri <spadubid@redhat.com>

This PR adds the placement API as defined here: kubevirt/hyperconverged-cluster-operator#718
to Node Labeller. 
Node Labeller pods are considered workload pods. The placement of the pods can be controlled by
spec.nodeSelector
spec.affinity
spec.tolerations

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

/cc @tiraboschi
Template Validator should only accept placement parameters from HCO_CR.spec.workloads

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/apis/kubevirt/v1/types.go
+++ b/pkg/apis/kubevirt/v1/types.go
@@ -116,10 +116,19 @@ type VersionSpec struct {
 	Version string `json:"version,omitempty"`
 }
 
-// Defines the configuration of the operand
+// Defines the configuration of the NodeLabeller
 type ComponentSpec struct {
-	// Defines the version of the operand
+	// Defines the version of the NodeLabeller
 	Version string `json:"version,omitempty"`
+
+	// Define the node affinity for NodeLabeller pods
+	Affinity v1.Affinity `json:"affinity,omitempty"`
+
+	// Define node selector labels for NodeLabeller pods
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Define tolerations for NodeLabeller pods
+	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 }
 
 // Defines the configuration of Template Validator
@@ -136,7 +145,7 @@ type TemplateValidatorSpec struct {
 	// Define node selector labels for TemplateValidator
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
-	// Define tolreations for TemplateValidator
+	// Define tolerations for TemplateValidator
 	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
 }
 
--- a/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
+++ b/roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
@@ -15,6 +15,9 @@ spec:
       labels:
         app: kubevirt-node-labeller
     spec:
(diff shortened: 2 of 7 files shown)


pkg/apis/kubevirt/v1/types.go
```diff
@@ -71,6 +71,9 @@ type VersionSpec struct {
 
 type ComponentSpec struct {
```
omeryahud, line 72: Please rename this struct to `NodeLabellerSpec` as it is not used by any other CR and looks confusing

pkg/apis/kubevirt/v1/types.go
```diff
@@ -71,6 +71,9 @@ type VersionSpec struct {
...
 	Version string `json:"version,omitempty"`
+	Affinity v1.Affinity `json:"affinity,omitempty"`
```
omeryahud, line 74: Please add comment descriptions above struct fields in order to not break the OpenAPI schema of the CRD.
See here for examples: https://github.com/kubevirt/kubevirt-ssp-operator/pull/225/commits/d6e77a2a07ecc0cc0ef36837c4390c9c186909f6#diff-5d05db759e7d0cd7129d4d2f1229d7afR127

omeryahud (changes requested):

pkg/apis/kubevirt/v1/types.go
omeryahud: @shwetaap You were correct offline, we shouldn't change the name of this struct

omeryahud:
@shwetaap Please look at this fix I posted https://github.com/kubevirt/kubevirt-ssp-operator/pull/231 and please apply it to your patch as well

pkg/apis/kubevirt/v1/types.go
shwetaap: Done

pkg/apis/kubevirt/v1/types.go
akrejcir: Why we shouldn't change it?

roles/KubevirtNodeLabeller/templates/kubevirt-node-labeller-ds.yaml.j2
```diff
@@ -15,6 +15,9 @@ spec:
       labels:
         app: kubevirt-node-labeller
     spec:
+      affinity: {{ cr_info.spec.affinity | default("", true) | to_json }}
```
omeryahud, line 18: Please don't forget to update default values, same as here https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/roles/KubevirtTemplateValidator/templates/service.yaml.j2#L61

shwetaap: Done

pkg/apis/kubevirt/v1/types.go
shwetaap: If a change is made here, then it must be updated in HCO as well. I am not sure if that is what we want to do
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/d4c0e35272c4ee7b05f0ecb84dc4cb2549643e09/pkg/controller/hyperconverged/hyperconverged_controller.go#L1307

akrejcir: No, we can leave it as is. I was just curious what was the reason.

PR #232, opened 2020-09-09
Fixing default values for placement fields of tempalte validator

omeryahud:
I missed this issue before because the test that should validate proper validator deployment was not yet enabled after OCP CI integration was complete. I enabled that test and also added a testcase to validate empty values for placement fields

Fixes default values for placement API

**Release note**:

```release-note
NONE
```


functests/10-test-template-validator-placement.sh
```diff
@@ -27,6 +29,8 @@ if [ "$DEPLOYMENT_FOUND" == "false" ]; then
   exit 1
 fi
 
+oc get -n ${TEST_NS} deploy virt-template-validator -ojson | jq '.spec.template.spec'
```
ksimon1, line 32: typo: `-ojson`

ksimon1 (approved):
/lgtm

functests/10-test-template-validator-placement.sh
```diff
@@ -27,6 +29,8 @@ if [ "$DEPLOYMENT_FOUND" == "false" ]; then
...
+
```
akrejcir, line 33: Why is this command here? Is it for logging?

functests/10-test-template-validator-placement.sh
```diff
@@ -48,6 +52,69 @@ if [ "$AFFINITY" != "testKey" ]; then
   RET=1
 fi
 
-oc delete -n ${TEST_NS} -f "${SCRIPTPATH}/10-template-validator-unversioned-cr.yaml" || exit 2
+oc delete -n ${TEST_NS} -f "${RES_DIR}/10-template-validator-affinity-nodeSelector-tolerations.yaml" || exit 2
+
+# Test if empty affinity/nodeSelector/tolerations values are propagated to the deployment
+oc create -n ${TEST_NS} -f "${RES_DIR}/10-template-validator-empty-affinity-nodeSelector-tolerations.yaml" || exit 2
```
akrejcir, line 58: I think it would be better to move this line after the loop, that waits for deployment delete. This way, it is probably a race condition.

akrejcir (changes requested):

functests/10-test-template-validator-placement.sh
omeryahud: Great catch, I agree

functests/10-test-template-validator-placement.sh
omeryahud: Yes, `test-runner.sh` hides it during `make functests`, and if the test fails, it will print it

PR #234, opened 2020-09-11
add missing version field in olm-catalog

ksimon1:
the .spec.customresourcedefinitions.owned[].version was missing in olm-catalog. This caused
that CVP tests were failing.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1878042

**Release note**:
```NONE

```

### DIFF
--- a/build/update-csv.py
+++ b/build/update-csv.py
@@ -58,6 +58,7 @@ _CRD_INFOS = {
                 'The version of the KubeVirt Template Validator to deploy',
             'displayName': 'Version',
             'path': 'version',
+            'version': 'v1',
             'x-descriptors': [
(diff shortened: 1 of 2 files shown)


build/update-csv.py
```diff
@@ -100,6 +102,7 @@
                 'The version of the aggregation rules to deploy',
             'displayName': 'Version',
             'path': 'version',
+            'version': 'v1',
```
tiraboschi, line 105: I think you miss it on line 90

tiraboschi (changes requested):

build/update-csv.py
ksimon1: fixed

tiraboschi (approved):

PR #239, opened 2020-09-17
Fix placement api

ksimon1:
replaced merge_type with apply in validator and labeller tasks
replaced jinja condition with normal condition

merge_type cause issue when user defined e.g. nodeSelector in CR and then removed it.
SSP operator was not able to remove it from deployment / daemonset.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1877698
**Release note**:
```
NONE
```

### DIFF
--- a/functests/10-test-template-validator-placement.sh
+++ b/functests/10-test-template-validator-placement.sh
@@ -52,6 +52,49 @@ if [ "$AFFINITY" != "testKey" ]; then
   RET=1
 fi
 
+timeout=300
+sample=10
+current_time=0
+
+echo "[test_id:4976]: Check if patch is applied without error"
+oc patch -n ${TEST_NS} KubevirtTemplateValidator kubevirt-template-validator --type='json' -p='[{"op": "remove", "path": "/spec/affinity"}, {"op": "remove", "path": "/spec/nodeSelector"}, {"op": "remove", "path": "/spec/tolerations"}]'
+#wait until operator applies newest configuration
+
+while [ $(oc get -n ${TEST_NS} pods | grep ^virt-template-validator.*Running | wc -l) -lt 2 ] ; do 
+  oc get pods
+  if [ $current_time -gt $timeout ]; then
+    RET=1
+    echo "template validator is not in running state"
+    break
+  fi
+  current_time=$((current_time + sample))
+  sleep $sample;
+done
+
+echo "[test_id:4977]: Check if nodeSelector is set as expected"
+NODE_SELECTOR=$(oc get -n ${TEST_NS} deploy virt-template-validator -ojson | jq '.spec.template.spec.nodeSelector' | tr -d '"')
+if [ "$NODE_SELECTOR" != "null" ] && [ "$NODE_SELECTOR" != "{}" ]; then
+  echo $NODE_SELECTOR
+  echo "template validator deployment is missing proper nodeSelector after update"
+  RET=1
+fi
+
+echo "[test_id:4978]: Check if tolerations is set as expected"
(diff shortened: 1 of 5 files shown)


functests/11-test-node-labeller-placement.sh
```diff
@@ -55,6 +55,31 @@ if [ "$AFFINITY" != "testKey" ]; then
   RET=1
 fi
 
+oc apply -n ${TEST_NS} -f "${RES_DIR}/11-node-labeller-empty-affinity-nodeSelector-tolerations.yaml" || exit 2
+#wait until operator applies newest configuration
+sleep 60
+
+NODE_SELECTOR=$(oc get -n ${TEST_NS} ds kubevirt-node-labeller -ojson | jq '.spec.template.spec.nodeSelector.testKey' | tr -d '"')
+if [ "$NODE_SELECTOR" == "testValue" ]; then
```
sarahbx, line 63: Where the YAML applied empties the spec values, we may want to check the result for a non-empty condition instead. If the entry is modified, but not deleted, that would still be a failure.

This would also apply to the two tests following this one.

functests/10-test-template-validator-placement.sh
```diff
@@ -52,6 +52,35 @@ if [ "$AFFINITY" != "testKey" ]; then
   RET=1
 fi
 
+echo "[test_id:4976]: Check if patch is applied without error"
+oc patch -n ${TEST_NS} KubevirtTemplateValidator kubevirt-template-validator --type='json' -p='[{"op": "remove", "path": "/spec/affinity"}, {"op": "remove", "path": "/spec/nodeSelector"}, {"op": "remove", "path": "/spec/tolerations"}]'
+#wait until operator applies newest configuration
+sleep 60
```
omeryahud, line 58: I think it is better to watch for a new `resourceVersion` on the CR instead of sleeping as this might be flakey

omeryahud (commented):
Have you tried updating the CR with `oc edit`? I remember this also having issues

functests/10-test-template-validator-placement.sh
```diff
@@ -52,6 +52,47 @@ if [ "$AFFINITY" != "testKey" ]; then
...
+
+echo "[test_id:4976]: Check if patch is applied without error"
+oc patch -n ${TEST_NS} KubevirtTemplateValidator kubevirt-template-validator --type='json' -p='[{"op": "remove", "path": "/spec/affinity"}, {"op": "remove", "path": "/spec/nodeSelector"}, {"op": "remove", "path": "/spec/tolerations"}]'
+#wait until operator applies newest configuration
+
+while [ $(oc get -n ${TEST_NS} pods | grep ^virt-template-validator.*Running | wc -l) -lt 2 ] ; do 
+  oc get pods
+  if [ $current_time -gt $timeout ]; then
+    RET=1
+    break
```
sarahbx, line 67: We should add a log statement here so we know that it exited due to timeout

roles/KubevirtTemplateValidator/tasks/main.yml
```diff
@@ -107,7 +107,7 @@
     namespace: "{{ cr_info.metadata.namespace }}"
     status:
       observedVersion: "{{ operator_version }}"
-  when: "{{ available==true }}"
+  when: "available==true"
```
omeryahud, line 110: I think the quotation marks are here by accident? I see you have removed them from similar statements

ksimon1: yes, you are right. Fixed

PR #249, opened 2020-10-12
fix generation of crds

ksimon1:
This fixes behaviour when command make generate adds in every iteration new
preserveUnknownFields field. When crd contains preserveUnknownFields,
it will not add a new one.

Fixes https://github.com/kubevirt/kubevirt-ssp-operator/issues/248

**Release note**:
```
NONE
```

### DIFF
--- a/hack/generate-crds.sh
+++ b/hack/generate-crds.sh
@@ -10,5 +10,7 @@ crd_files=$(ls deploy/crds/*crd.yaml)
 
 for crd in $crd_files
 do
-  echo "  preserveUnknownFields: false" >> $crd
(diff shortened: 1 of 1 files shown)


hack/generate-crds.sh
```diff
@@ -10,5 +10,7 @@ crd_files=$(ls deploy/crds/*crd.yaml)
 
 for crd in $crd_files
```
nunnatsa, line 11: maybe also do something more robust, like 
```bash
SELF=$( realpath $0 )
BASEPATH=$( dirname $SELF )
./operator-sdk generate crds

for crd in deploy/crds/*crd.yaml
do
  ...

```

PR #255, opened 2020-12-07
Report available condition regardless of which template versions were deployed

omeryahud:
Previously, all templates were aligned to the same version, but a previous fix (https://github.com/kubevirt/kubevirt-ssp-operator/pull/253)

locked deprecated templates to older versions, so the operator did not report the Available condition.

This fix makes the operator report Available when all templates were deployed, regardless of their version.

**Which issue(s) this PR fixes** 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1904830

**Release note**:

```release-note
None
```

### DIFF
--- a/roles/KubevirtCommonTemplatesBundle/tasks/main.yml
+++ b/roles/KubevirtCommonTemplatesBundle/tasks/main.yml
@@ -55,12 +55,12 @@
 # If the old cr exists, it would have a metadata field
 - name: Check if the old CR exists
   set_fact:
-    old_cr_exists: "{{ true if (old_cr is defined) and (old_cr.metadata | default(None)) != None else false }}"
+    old_cr_exists: "{{ True if (old_cr is defined) and (old_cr.metadata | default(None)) != None else False }}"
 
 - name: Filter for templates owned by the old cr
   set_fact:
     old_cr_templates: "{{ templates | k8s_owned_by(old_cr) }}"
-  when: "{{ old_cr_exists==true }}"
+  when: old_cr_exists == True
 
 # Inject ownerReferences
 - name: Inject owner references for KubevirtCommonTemplatesBundle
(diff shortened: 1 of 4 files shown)


roles/KubevirtCommonTemplatesBundle/tasks/main.yml
```diff
@@ -69,23 +69,19 @@
...
-
 - name: "Get all templates"
   set_fact:
-    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.result.results[0].result.apiVersion, kind='template', label_selector=label) | length }}"
+    deployed_templates_after: "{{ lookup('k8s', api_version=ct_status.result.results[0].result.apiVersion, kind='template', label_selector='template.kubevirt.io/type=base') | length }}"
 
 - name: "Set Available status"
   set_fact:
-    available: "{{ true if new_templates <= deployed_templates_after else false }}"
+    available: "{{ True if new_templates <= deployed_templates_after else False }}"
```
sarahbx, line 84: Ansible doesn't particularly care about True vs true, but yaml linter may. Why the change?

omeryahud: Just for consistency really, nothing more

omeryahud:
I've updated the templates deployment tests to wait for the condition (should have done it a long time ago, but missed it)

omeryahud:
Functests are failing due to a flakey test I am addressing the issue in a different PR https://github.com/kubevirt/kubevirt-ssp-operator/pull/256

roles/KubevirtCommonTemplatesBundle/tasks/main.yml
dankenigsberg: for a layperson like myself, it is important to keep cosmetics like this on another PR.

PR #259, opened 2020-12-10
updated common templates to version v0.13.0-rc.0

MarSik:
updated common templates to version v0.13.0-rc.0


omeryahud:
This templates release should not be merged to this operator, only to the new one
/close

PR #261, opened 2021-01-05
updated common templates to version v0.13.0

MarSik:
updated common templates to version v0.13.0


omeryahud:
Irrelevant for this operator
/close

